PR #119, opened 2018-07-23
AnnotationModule v0

pjl54:


config.ini
```diff
@@ -1,181 +1,186 @@
...
+image_work_size = 1.25x
+
+#three options: relative2mask, absolute, relative2image
+mask_statistics = relative2mask
+
+[BasicModule.getBasicStats]
+image_work_size = 1.25x
+
+[BasicModule.getMag]
+confirm: False
```
choosehappy: you should remove these since in the latest version this has been pushed into Base Image

config.ini
```diff
@@ -1,181 +1,186 @@
...
+
+[AnnotationModule.maskOutAnnotation]
+#annotation_path = []
```
choosehappy: its okay to have this commented out if there is a reasonable default in your code, but i prefer to have it commented out *with* the default value stored here, so users can understand what the expected behavior is

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
...
+def maskOutAnnotation(s,params):
+
+    annotationPath = params.get("annotation_path",s["dir"])
+    annotationPath = annotationPath + os.sep + s["filename"].rsplit(".",1)[0] + ".xml"
+
+    points = getPoints(annotationPath)
+
+    downsize_factor = int(s["height"])/s["img_mask_use"].shape[0]
+
+    points = [(a[0][0] / downsize_factor, a[0][1] / downsize_factor) for a in zip(points)]
```
choosehappy: nicely done

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
...
+
+    img = Image.new('L', (s["img_mask_use"].shape[1], s["img_mask_use"].shape[0]), 0)
```
choosehappy: is PIL already a HistoQC requirement, or does this approach require adding on another dependency?

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
...
+import cv2
+import numpy as np
+import openslide
+import os
+from skimage import io
+from BaseImage import printMaskHelper
+from PIL import Image, ImageDraw
+
+def getPoints(xmlfile):
+    # create element tree object
```
choosehappy: try to think about other users. is it obvious from your code what this XML file should contain? what format it should be in? what if someone wants to write their own XML file programmatically and have it ingested by your module, should they try and reverse engineer it from your parser?

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
+import csv
+import xml.etree.ElementTree as ET
+import cv2
+import numpy as np
+import openslide
+import os
+from skimage import io
+from BaseImage import printMaskHelper
+from PIL import Image, ImageDraw
+
```
choosehappy: your module has essentially at least doubled the # of dependencies that HistoQC requires. this is non-obvious, and many of these are not installed by default (e.g., xml). while i would prefer a leaner version, if this is what has to happen, then so be it. since the modules themselves are loaded dynamically when included in the pipeline, users which don't use your module won't have these requirements. as a result, when adding dependencies you need to modify the documentation (e.g., Readme.md in this case) to state what they are. In this case, i would add a sub section below the current one with the name of your module and state its unique additional dependenices, so users who don't need it won't need to meet the requirements

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
...
+    # create empty list for annotations
+    annotations = []
+
+    for annotation in root.findall('Annotation'):
+        for regions in annotation.findall('Regions'):
+            for region in regions.findall('Region'):
+                for vertices in region.findall('Vertices'):
+                    points[annCount] = [None] * len(vertices.findall('Vertex'))
+                    k = 0
+                    for vertex in vertices.findall('Vertex'):
```
choosehappy: can you use "enumerate" here instead of the k counter?

AnnotationModule.py
choosehappy: though you may want to use "p" for "points" instead of "a" for...unknown?

BaseImage.py
```diff
@@ -1,134 +1,134 @@
...
-
-                output = np.concatenate(output, axis=1)
-                output = output[0:round(dim_base[1] * 1 / down_factor), 0:round(dim_base[0] * 1 / down_factor), :]
-                self[key] = output
-            else:
-                logging.error(
-                    f"{self['filename']}: Unknown image level setting: {dim}!")
-                return -1
-        return self[key]
+import logging
```
choosehappy: try to not submit white space changes, makes it difficult to see the true changes. there is likely an option in the UI that will allow this, but for this commit, this may help:

https://stackoverflow.com/questions/3515597/add-only-non-whitespace-changes

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
...
+    for pointSet in points:
+        ImageDraw.Draw(img).polygon(pointSet, outline=1, fill=1)
+
+    mask = np.array(img)
+    io.imsave(s["outdir"] + os.sep + s["filename"] + "_annotation.png", mask * 255)
+    
+    prev_mask = s["img_mask_use"]
+    mask = s["img_mask_use"] & (mask > 0)
+    s["img_mask_annotation"] = (mask * 255) > 0
+    s["img_mask_use"] = s["img_mask_use"] & mask
```
choosehappy: prev_mask line should be right before this one

AnnotationModule.py
```diff
@@ -0,0 +1,63 @@
...
+
+    points = [(a[0][0] / downsize_factor, a[0][1] / downsize_factor) for a in zip(points)]
+
+    img = Image.new('L', (s["img_mask_use"].shape[1], s["img_mask_use"].shape[0]), 0)
+
+    for pointSet in points:
+        ImageDraw.Draw(img).polygon(pointSet, outline=1, fill=1)
+
+    mask = np.array(img)
+    io.imsave(s["outdir"] + os.sep + s["filename"] + "_annotation.png", mask * 255)
```
choosehappy: do you want to save the annotation mask, or the difference between the previous img_make_use and the new img_mask_use or both?

choosehappy (changes requested):
looks like you're missing a change to the UI file which will add in your thumbnail to the front end?

AnnotationModule.py
pjl54: I think I want to save the annotation mask. Does it make sense to do that?

AnnotationModule.py
pjl54: Sure, as soon as I learn how to use enumerate.

AnnotationModule.py
pjl54: Sure

AnnotationModule.py
pjl54: PIL is a new dependency.

config.ini
pjl54: The default value is the directory that the images are in. Is there a way to communicate that in the config file?

choosehappy: tricky. just put a note above the comment to explain, i don't there is a notation which will be obvious in this case. also mention the naming convention, which i assume is base_filename_image.svs - >base_filename_image.xml

AnnotationModule.py
choosehappy: its up to you : ) whatever you think is most useful. you can also opt to save the fused version of the two masks. if you're happy with the results, then likely others will be, but given the open source nature of histoqc we'll be able to get feedback if they're not

PR #130, opened 2018-11-14
Fix default value when getting from param dict in getMag

larsjr:
Fix the default value when getting "confirm_base_mag" key from the params dict. Avoids error in strtobool function.


choosehappy:
yup...missed that one...thanks

PR #133, opened 2019-05-06
Create requirements.txt

birm:
This way it may be easier to recommend ```pip install -r requirements.txt``` instead of having user collect the python requirements themselves.


choosehappy:
you're absolutely right, it was somewhere on my to-do list but didn't get around to it

thanks for this!

PR #139, opened 2019-07-24
Update BrightContrastModule.py

Rexmiao:
Tenen Grad -1

### DIFF
--- a/BrightContrastModule.py
+++ b/BrightContrastModule.py
@@ -1,5 +1,6 @@
 import logging
 import numpy as np
+from skimage.filters import filters
 from skimage.color import convert_colorspace, rgb2gray
 from distutils.util import strtobool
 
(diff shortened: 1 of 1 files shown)


choosehappy:
Did you test this? are you sure it works correctly? see my comment in the code

BrightContrastModule.py
```diff
@@ -56,4 +57,10 @@ def getContrast(s,params):
     #RMS contrast
     rms_contrast=np.sqrt(pow(img - img.mean(), 2).sum() / img.size)
     s.addToPrintList("rms_contrast", str(rms_contrast))
+         
+    #TenenGrad contrast
+    tmp = sobel(img)
```
choosehappy, line 62: looking at this sobel(img)

doesn't this assume that img is a matrix?

Above, if limit_to_mask is true:

```
    if (limit_to_mask):
        img = img[s["img_mask_use"]]
```

isn't img then a vector, as its been indexed by a binary mask matrix, and not a 2d array/matrix?

PR #143, opened 2019-11-27
Workaround for fixing UTF8 issues

petrovm3:
Problem related to UnicodeDecodeError : ascii codec can't decode byte.
Solved by adding a few docker lines the lines above.

https://stackoverflow.com/questions/47013840/docker-compose-up-giving-error-unicodedecodeerror-ascii-codec-cant-decode-by

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,25 @@
 FROM ubuntu
 
-RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


choosehappy (approved):
looks great, thanks for the contribution!

PR #175, opened 2020-05-05
add annotation Module

pjl54:


AnnotationModule.py
```diff
@@ -0,0 +1,81 @@
...
+    # get root element
+    root = tree.getroot()
+
+    points = []
+
+    for annotation in root.findall('Annotation'):
+        for regions in annotation.findall('Regions'):
+            for region in regions.findall('Region'):
+                for vertices in region.findall('Vertices'):
+                    points.append([None] * len(vertices.findall('Vertex')))
```
choosehappy: seems like this section should be done with list compression, similar to lines 33 below

AnnotationModule.py
```diff
@@ -0,0 +1,81 @@
...
+    points = get_points(xml_fname)
+
+    resize_factor = np.shape(s["img_mask_use"])[1] / s["image_base_size"][0]
+
+    points = resize_points(points, resize_factor)
+
+    mask = np.zeros((np.shape(s["img_mask_use"])[0],np.shape(s["img_mask_use"])[1]),dtype=np.uint8)
+
+    for pointSet in points:
+        cv2.fillPoly(mask, [np.asarray(pointSet).reshape((-1, 1, 2))], 1)
```
choosehappy: can you use skimage stead of cv2. we already have it as a dependency while cv2 is not

example is here: https://scikit-image.org/docs/0.7.0/auto_examples/plot_shapes.html

AnnotationModule.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return mask
+
+def xmlMask(s,params):
+    logging.info(f"{s['filename']} - \txmlMask")
+    mask = s["img_mask_use"]
+
+    xml_basepath = params.get("xml_filepath","")
```
choosehappy: would suggest having the default be absent, which i believe will return None, then later on  line 59 becomes "if not xml_basepath"

which is more pythonic

AnnotationModule.py
```diff
@@ -0,0 +1,81 @@
...
+    xml_suffix = params.get("xml_suffix", "")
+    if len(xml_basepath)==0:
+        xml_basepath = s["dir"]
+
+    xml_fname = xml_basepath + os.sep + s['filename'][:s['filename'].rindex('.')] + xml_suffix + '.xml'
+
+    logging.info(f"{s['filename']} - \tusing {xml_fname}")
+
+    annotationMask = mask_out_annotation(s,xml_fname) > 0
+    io.imsave(s["outdir"] + os.sep + s["filename"] + "_xmlMask.png", img_as_ubyte(annotationMask))
```
choosehappy, line 88: use:
https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f

pathlib

BaseImage.py
```diff
@@ -58,8 +58,10 @@ def __init__(self, fname, fname_outdir, params):
         self["dir"] = os.path.dirname(fname)
 
         self["os_handle"] = openslide.OpenSlide(fname)
+        self["image_base_size"] = self["os_handle"].dimensions
         self["image_work_size"] = params.get("image_work_size", "1.25x")
         self["mask_statistics"] = params.get("mask_statistics", "relative2mask")
+        self["base_mag"] = getMag(self, params)
```
choosehappy, line 64: you shouldn't have to call getMag twice, especially since this may be an expensive function in the future. also, if folks write buggy code and you end up with 2 diference base_mag values you're life gets complicated

instead keep line 64, and use the resulting "S" value for like 65

choosehappy (changes requested):
a few minor house keeping issues in the code

one important component missing...you never mention how/where/what these xml files look like. how are they generated?  you can add a lot of this information in as a comment at the top of the module

can you also create a directory called "AnnotationModule", you can move your config file in there, and provide a full example of an xml file in there. i think that should nicely cover documentation needs

PR #179, opened 2020-06-04
Support tsv input with multiple filenames per line

pjl54:


choosehappy:
i don't think this works as intended. the goal is to load a histoqc results.tsv file and reprocess it (after e.g., filtered using the front end)

the first column is a filename, thus the .split("\t")[0] to get the filename

this looks like it assumes multiple files on the same line. when in fact it should be 1 file per line

if you have a tab seperated list of files, you can change it to one per line using sed, something like this (unchecked):
 
```
sed 's/\t/\n/g' inputfile.tsv
```

PR #180, opened 2020-06-14
GeoJSON annotations for AnnotationModule

jjhbw:
Hi! Great fan of this project.

I wanted to limit my QC analyses to certain ROIs that were annotated in QuPath. Exporting to the ImageScope XML format from QuPath turned out to be kind of a hacky affair because of a lack of built-in support. However, QuPath's GeoJSON exports are great and AFAIK up to spec.

I have extended the AnnotationModule a bit so it supports imports of GeoJSON annotations directly, without first converting to ImageScope XML. I have tested it on a few complicated examples I had lying around, and it seems to work ok.

Let me know what you think!


choosehappy:
thanks so much for your efforts! Very interested in having these functionalities

@pjl54 can you do the first review since you're the original creator of the annotation module?

AnnotationModule.py
```diff
@@ -48,11 +47,33 @@ def get_points_from_xml(xml_fname):
...
+    point_sets = []
+    for annot in geojson:
+        geometry = annot['geometry']
+        geom_type = geometry['type']
+        coordinates = geometry['coordinates']
+        if geom_type == 'MultiPolygon':
+            for roi in coordinates:
+                for points in roi:
+                    point_sets.append([(coord[0], coord[1]) for coord in points])
+        elif geom_type == 'Polygon':
```
pjl54, line 67: Suggest changing to 

elif geom_type == 'Polygon' or geom_type == 'LineString':

As LineString is the geometry type produced by the magic wand tool and has the same field names as Polygon.

AnnotationModule.py
```diff
@@ -99,4 +120,35 @@ def xmlMask(s,params):
...
+def geoJSONMask(s, params):
+    logging.info(f"{s['filename']} - \tgeoJSONMask")
+    mask = s["img_mask_use"]
+
+    geojson_basepath = params.get("geojson_filepath",None)
+    geojson_suffix = params.get("geojson_suffix", "")
+    if not geojson_basepath:
+        geojson_basepath = s["dir"]
+
+    fname = geojson_basepath + os.sep + PurePosixPath(s['filename']).stem + geojson_suffix + '.json'
```
pjl54, line 143: The configuration reads: "leave blank to replacing image filename extension with "geojson""

But here the image extension it looks for is 'json'.

AnnotationModule.py
pjl54: Whoops, they're the same field names but slightly different formats. I suggest adding the following to allow use of magic wand annotations:

        elif geom_type == 'LineString':            
            point_sets.append([(coord[0], coord[1]) for coord in coordinates])

pjl54 (changes requested):

pjl54:
This is a great addition to AnnotationModule, thanks for your work. I had a few small changes but otherwise it looks good and works in my testing. 

The json annotations exported by Qupath handle non-contiguous annotations better than the exported Imagescope XMLs do. I definitely agree with using the json files whenever possible.

One note, in my testing this didn't respect holes in annotations, though that may have to do with the script I used for exporting annotations from Qupath.

AnnotationModule.py
jjhbw: Ah, interesting! I thought LineString's were just simple lines? I thought the QuPath magic wand tool also produced ROIs with an actual area. How can LineStrings be used to represent these areas? Are these lines simply joined at the ends?

If thats the case, seems like an easy enough addition! Good idea.

AnnotationModule.py
jjhbw: Good find! Will fix.

jjhbw:
> in my testing this didn't respect holes in annotations

Hah! This happened to be exactly what my team just noticed. We're heavy users of 'de-annotations' i.e. holes to exclude artefacts and the like. 

I think this is the code to look at for that behaviour:
https://github.com/choosehappy/HistoQC/blob/0a747226fb34d1384d6fe74183c49167febb4faf/AnnotationModule.py#L67-L71

It looks like this function implicitly takes the union of all ROIs to make the mask. This is the simplest approach, but obviously does not work with holes...

Nested ROIs come with a few challenging mindbenders like dealing with multiple levels of nesting (think donut-in-donut-like shapes). Python has a nice ecosystem for dealing with Shapes (big fan of Shapely). I'll take a run at it in a separate PR.

AnnotationModule.py
jjhbw: Added in bc83c5feb1e1fb37fd208cd7cccd686f33fe9c1d 👍

jjhbw:
Also, raising an exception from `get_points_from_geojson` when the GeoJSON contains a geometry type that's not supported may be a bit too drastic. I can imagine that people just export all their annotations to GeoJSON from QuPath, including maybe Point types. If we raise an exception when encountering these types, the user will have to filter/preprocess their GeoJSON's first, which seems pointless.

Maybe we can simply ignore these types instead, or log a warning?

AnnotationModule.py
```diff
@@ -99,4 +122,35 @@ def xmlMask(s,params):
...
```
choosehappy, line 143: need to have an error check here to catch if the filename doesn't exist. should simply continue + post a warning to the user

i imagine some cases, when running at scale, not 100% of images will have annotations

choosehappy: it looks like this check also needs to be done in @pjl54  's version as well, would you mind adding?

choosehappy (commented):
looks great! one small point to address

choosehappy:
> Maybe we can simply ignore these types instead, or log a warning?

yup, a warning is appropriate in this case


Regarding: https://github.com/choosehappy/HistoQC/blob/0a747226fb34d1384d6fe74183c49167febb4faf/AnnotationModule.py#L70

Yes, currently we're only supporting positive annotations, simply because negative ones as you point out can become quite complex. 

if this needs to be modified we should likely make some type of an annotationROi class, which has the target value {0,1} and the points themselves, then this loop here would change minimally, and delegate the complicated negative component to a different function

Thoughts?

pjl54:
Imagescope XMLs have a "negative ROI" property that is used for holes. It seems it's possible to get the holes separately in Qupath annotations too: https://github.com/qupath/qupath/issues/219 

You could then do:

`mask[rr,cc] = not negative_flag`

I think that would work as long as all the negative annotations were at the end of the list.

choosehappy:
you'd still need to wrap into a high-order structure.

the "points" variable creates multiple polygons, one needs to keep track of which should be positive and which should be negative

it would be more like

```
for polyobj in polyobjs:
        poly = np.asarray(polyobj.pointSet)
        rr, cc = polygon(poly[:,1],poly[:,0],mask.shape)
        mask[rr,cc] = polyobj.val

```

AnnotationModule.py
jjhbw: Good idea. I have added this in 9e833cda5af7018632c29607a46202465f670d67

jjhbw:
Made another minor change: when an incompatible geometry type (like a Point) is found in the geojson file, it now simply logs a warning and moves on (d7cb3f5cd0286211f5fd52d035abe65cf87c87fc).

The 'negative ROI' issue is best suited for a separate PR I think. I think there's quite a bit of tricky logic required to make this work correctly. I think our best bet is using a library like Shapely to handle all the geometry parsing. Turning complex, nested geometries into raster masks should also be possible with it.

choosehappy:
looks good! merged in! thanks for your effort!

i agree,  the negative ROI component can become a bit tricky, best set out in a separate PR.  we'll need to create some test data to try out various use-cases, adding some complexity on the verification step

PR #182, opened 2020-07-20
Added UMAP Module

foggydae:
The UMAP functionality has been added and wrapped up into a single script, which can be easily converted into an object-oriented class and embedded into other systems. 

The front end code has also been restructured to be more modular. This is a work in progress and will be addressed in another branch starting from now.


pjl54:
Are features normalized prior to UMAP embedding? I couldn't find any normalization in this code, but I may have missed it.

choosehappy:
UMAP doesn't use normalization to my knowledge:

"This provides a very efficient approximate stochastic gradient descent
algorithm since there is no normalization requirement."

"This is largely due to the fact that UMAP does not require global
normalisation  "

https://arxiv.org/pdf/1802.03426.pdf

Thoughts?

pjl54:
I don't think UMAP requires normalization, in the sense that it will produce an embedding regardless of any normalization, but when the features are different orders of magnitude, normalization makes the embedding more meaningful. I think Euclidean distance is the default here and in most UMAP packages, which is only meaningful if the features are on the same scale. 

From the UMAP docs written by the first author of that paper: https://umap-learn.readthedocs.io/en/latest/faq.html

choosehappy:
as a first pass, we've tested it with a few datasets and the histoqc features appear to be "embedable" in their current scales and already reveal when notable batch effects are present, this is likely because the *default* embedding dimensions are of approximately of the same magnitude.

otherwise, "blind" normalization can be a bit tricky and will require more thought and evaluation. for example, the MSE value likely shouldn't ever be normalized, as it is unbounded. the artifact features may or may not be [0,1] depending on the user config file (relative mask, pixel level, etc)

it feels a normalized approach would involve a lot more moving parts and testing before we could land on something we would consider a "best practice". in absence of a validated best-practice, i'm inclined to leave things unnormalized since this is likely somewhat more intuitive at first pass (larger numbers --> larger differences -- > larger impact) than doing something unvalidated behind the scenes to break that property.

we'll need to pay close attention to our experiences to determine how best to move forward from this point

Issue #181, opened 2020-07-17
'Number of labels=34788393 does not match number of samples=11596131' when training pen classifier


markemus:
I'm trying to create a mask for the pen classifier module, but I'm getting this error during the training step. Here is the training image and mask: https://imgur.com/a/HwaIaB2

Both the image and mask have the same x,y shape.

Rexmiao:
Hi this is Runtian, I am happy to help with HistoQC. But I am kind of confused by your questions. My email is **rxm723@case.edu**.
Would you like to send me more details through email? The screenshot of trace error will be very helpful. Thanks

markemus:
Thanks @Rexmiao I emailed you.

markemus:
I was able to fix this bug by changing ClassificationModule.py:181 from:

`mask = io.imread(ex[1]).reshape(-1, 1)`

to:

`mask = io.imread(ex[1], as_gray=True).reshape(-1, 1)`

skimage.io.imread() reads images inconsistently based on unseen metadata. cv2.imread() is more predictable, just saying :)

choosehappy:
great, glad you figured it out!

agree, cv2 is a better (and faster!), but in this context represented
another "heavy" dependency, so we opted to use the existing dependencies

in this case, seems like just adding the option flag would fix the issue
for everyone, @Runtian Miao <rxm723@case.edu>  can you modify as needed?

PR #183, opened 2020-07-23
Add the option flag (Classification Module) #181

Rexmiao:
Add the option flag, since skimage.io.imread() reads images inconsistently based on unseen metadata

#181


choosehappy:
thanks!

PR #186, opened 2021-02-18
python installable histoqc

ap--:
Hi @choosehappy,

Thanks for creating this excellent tool! I've been playing around with some of the HistoQC functionality and found it hard to use outside of its intended use case via `qc_pipeline.py`

So this PR is basically turning HistoQC into a python package that is pip installable, and setuptools_scm versioned.
`qc_pipeline.py` is now provided via either calling `python -m histoqc` or explicitly as a console_script entrypoint `histoqc_pipeline`

To not break every possible workflow there's things that would need to be considered:
- the default config relies on a few files present and reachable from the repository root. I suggest this would need to be provided as package data if you consider the default config use case to be useful
- the symlinking to provide an easy way to display the UI with the created dataset is broken as of now. This could probably be implemented as another console_script entrypoint that would just start a webserver with the UI pointing to the provided location
- the `error.log` logging handler should be initialized later

Consider this PR as a starting point for discussion, and let me know if you are interested in making histoqc pip installable or if you are opposed to the idea.

Cheers,
Andreas 😃


choosehappy:
Apologies for the delay in responding to this amazing work! 

I’ve been thinking about this deeply, and also had a chance to review all of the code in your pull request

Thanks for doing the packaging, its unfortunately not something I have extensive experience with. In looking at the base-code (i.e. non-packaging code) changes, they seem fairly straightforward

One thing that we do need is to have version numbers for all the dependencies. From a quality control, reproducibility standpoint, you really want to have those pinned such that when executing the same pipeline on the same images over time, one can expect the same results. I think that is a rather minor point though, and I suspect and hope you’ll agree quite readily with it : )


My major concerns are inline with the 3 points you’ve called out for consideration
1)	Users will need to have access to the set of provided config files in order to use those as basis for their own. The default configs are used by many people as the first line approach, from which they can make modifications to tailor to their use-case. 

That said, a better approach may be something like:

Python -m histoqc --generate-config [config-type]

Which pulls out the template from the package data into the current user directory, and then acts as a basis to begin

2)	Error.log, this has in fact stumped me for quite a while, and if you notice it doesn’t work as intended. I’ve not found a way to “shutdown” the logger, such that the last line of the histoqc pipeline should be to move the error log into the associated output directory. This doesn’t happen, and as a result it in fact keeps getting appended to overtime and can become quite large and not interesting to use. I wonder if initializing it later would result in this problem being resolved automatically

3)	UI, this seems like the most critical component to resolve, and its not clear to me how that would take place if all of the needed files are in the package data directory. This was written specifically to avoid having to install a webserver (for a number of reasons), and to enable a simple double click on the index file from the file browser. Now I’m wondering if the inbuilt http.server in python would be useful here (or if that is even a good idea), but I can’t see how to connect different data directories to the html file. 

Any thoughts?

ap--:
Sorry for the delay on my side, too. I was super busy last week 😃 

Regarding your comments:

### pinned versions

> One thing that we do need is to have version numbers for all the dependencies. From a quality control, reproducibility standpoint, you really want to have those pinned such that when executing the same pipeline on the same images over time, one can expect the same results. I think that is a rather minor point though, and I suspect and hope you’ll agree quite readily with it : )

I fully agree! For the specific use case for which I am currently trying to integrate histoqc, we manage the environment including pinned versions of all packages ourselves, which is why I unpinned the dependency versions in the setup.cfg file.
(we need newer versions of some of the dependencies in our parent environment, which would otherwise cause conflicts)
I think there's an option to accommodate both scenarios:

Using `setuptools`'s `extras_require` option we could make

```sh
# the following would install histoqc and download the newest versions of the dependencies
$> pip install histoqc

# the following would pin the versions to the ones used for current histoqc
# (I assume those are the ones used for the publication?)
$> pip install histoqc[pinned]

# note: `pinned` could be renamed to whatever, and there could be multiple different sets of pinned versions
$> pip install histoqc[published]
$> pip install histoqc[v2021]
# ... etc etc
```

Other options in case you're worried that most users will just type `pip install histoqc` and accidentally install the unpinned version, would be:
- add a commandline option that would unpin the dependencies
  ```$> pip install --unpin-histoqc-dependencies histoqc```
  
- or make `pip install histoqc` require the user to pick an extra. So only `pip install histoqc[published]` will lead to a working install. And `pip install histoqc[unpinned]` would allow for our usecase. 

Let me know what you prefer and I'll set it up.

### config templates

> That said, a better approach may be something like:
> `python -m histoqc --generate-config [config-type]`
> Which pulls out the template from the package data into the current user directory, and then acts as a basis to begin

That's easy to do. I'll set that up and include it in the PR.

### error logging

> Error.log, this has in fact stumped me for quite a while, and if you notice it doesn’t work as intended. I’ve not found a way to “shutdown” the logger, such that the last line of the histoqc pipeline should be to move the error log into the associated output directory. This doesn’t happen, and as a result it in fact keeps getting appended to overtime and can become quite large and not interesting to use. I wonder if initializing it later would result in this problem being resolved automatically

The python logging module is pretty confusing and getting it right can be a bit tricky. Can you describe the exact intended behavior to me? I should then be able to make it work. So as I see it right now: it should log to console and the logfile, and move the file to the output dir when the pipeline finished/crashed?

### user interface

> UI, this seems like the most critical component to resolve, and its not clear to me how that would take place if all of the needed files are in the package data directory. This was written specifically to avoid having to install a webserver (for a number of reasons), and to enable a simple double click on the index file from the file browser. Now I’m wondering if the inbuilt http.server in python would be useful here (or if that is even a good idea), but I can’t see how to connect different data directories to the html file.

I would have added a console script that would spawn a python webserver (`python -m http.server`) that you point to the output directory.
So something like
```
$> python -m histoqc.ui /path/to/output_dir
# or as a console entry point
$> histoqc_ui --port 8080 /path/to/outdir
spawning histoqc user interface at http://127.0.0.1:8080 ...
```

We could also support some way of extracting the ui folder structure to an arbitrary location
```sh
$> histoqc_ui --deploy /some/folder
extracting static user interface ...
```
That would support the current workflow.


### how to continue

Let me know how you would prefer to continue iterating on these issues.
I could split up the issues into separate PRs or I could keep pushing to this PR until all issues are resolved.

Have a great week 😃 
Andreas


_P.S.: FYI, I also sketched out a pep8 conform and pep484 typed python api for histoqc that allows easier integration into other codebases. This also made it easier to support testing frameworks (in this case pytest) and code coverage analysis, which might be of interest when considering the software side of the whole quality control issue:_ https://github.com/bayer-science-for-a-better-life/HistoQC/commit/e2303362163864fe41bb2e51eacd429329912ffb

_P.P.S.: Once histoqc is a python package, I can also offer some assistance in case you want to automatically deploy histoqc to pypi for every release created in the GH repository._

choosehappy:
Very informative, thank you!

In regards to pinning, I have a question that you might be able to help me with

The version of histoqc in the master currently requires a specific maximum version of skimage, because in later versions of it they broke backward compatibility by changing the function prototype for one of the filters we use.  

This is actually one of the main headaches I'm facing in regards to versioning + pinning, and I see you encountered it here:

https://github.com/bayer-science-for-a-better-life/HistoQC/commit/12931262da0c1e54b44088222bcbca98fca5b790

That said, maybe it makes sense to simply freeze that into a definitive tagged-release, and then push all the library versions to the most recent ones with the associated needed code fixes, and begin from that point to do the packaging




**Logging**

> it should log to console and the logfile, and move the file to the output dir when the pipeline finished/crashed?

Yes, exactly what I was thinking. Does that make sense to you? Since the log is _specific_ to that output, it seems like it should be bundled together in the same place

Unfortunately, currently the log file remains open, which prevents it from being moved to the output directory. When running histoqc again, it open the log in append mode and keeps adding to it, which gives a full-recounting of all files ever processed, but this doesn't seem too useful to me


**UI**

This is still a bit unclear to me, since the UI files (index.html + js libraries) and the output directory from the histoqc run would be located in two different places, and for the UI to work, the histoqc output needs to be in the relative .\Data sub-directory from the index.html

I resolve this issue in Linux by linking the directory. In Windows I was manually copying the histoqc output directory to the .\Data directory in UserInterface since it doesn't support linking in an attractive way. I've not really found an easier way of doing that. 

Starting http.server also doesn't seem to resolve this issue, unless I'm missing some behind the scenes work? Thoughts?


**To continue**

> I could split up the issues into separate PRs or I could keep pushing to this PR until all issues are resolved.

my preference would be to put it into this PR and then on our side we can do a single pull, test, and merge when everything is in place

** pep8 / pep484 **

Some real heavy lifting here! If you're interested and think it will also be useful to others, I'm happy to pull in these additions as well. A quick review seems to indicate to me its all value-add without any breaking any functionality, so more than welcome!


Thanks again!

ap--:
#### dependency pinning

> The version of histoqc in the master currently requires a specific maximum version of skimage, because in later versions of it they broke backward compatibility by changing the function prototype for one of the filters we use.
>
> This is actually one of the main headaches I'm facing in regards to versioning + pinning, and I see you encountered it here:
>
> bayer-science-for-a-better-life@1293126
>
> That said, maybe it makes sense to simply freeze that into a definitive tagged-release, and then push all the library versions to the most recent ones with the associated needed code fixes, and begin from that point to do the packaging

Handling breaking upstream changes can often be a bit annoying. There's multiple ways this could be handled, one of which is what you propose:
- tag the current master to `v1.0.1` and restrict the version of `scikit-image<0.16.1`
- and then introduce the change and create a new release of histoqc `v1.1.0` that depends on a minimum version of `scikit-image>=0.16.1`

For sticking to good versioning conventions I would recommend following https://semver.org
The gist of semver is that once you reach major version 1, every time you introduce a breaking change in your code, you have to increase your major version. That makes it easy for software that depends on your code to pin the major version, and still benefit from bugfixes. Newly introduced features bump the minor version, and pure bugfixes increase the micro (or patch version). Sadly in reality many libraries don't stick to that and it sometimes happens that breaking changes get introduced in between. But it's still good practice. There's then also plenty of room for debating if a change requires to bump major, minor, or patch, which is why some software never goes to v1 (https://0ver.org/) 😄 

Another option is often used if the actual interfacing code with the dependency is very slim. So actually the use of a few functions of scikit image qualifies in this case. That option boils down to checking the version of scikit-image in the HistoQC code and then call `remove_small_holes` with the correct arguments for that version.

A good way of detecting issues with upstream code is running tests on a CI (so for example Github Actions) periodically (maybe every week or so). The test will then fail in case a dependency introduces an unexpected change.


#### logging

> Yes, exactly what I was thinking. Does that make sense to you? Since the log is specific to that output, it seems like it should be bundled together in the same place

Makes perfect sense. Is there any reason why the log can't be written directly to the output folder? Is the use case here, that the log file is available from the current working directory? 


#### user interface

> Starting http.server also doesn't seem to resolve this issue, unless I'm missing some behind the scenes work? Thoughts?

Yes, sorry. You're of course right that the default http.server won't cut it in this case. But it's easy to derive your own subclass and mount the output dir under `localhost:port/Data` and serve the UI from a different location. 


#### pr workflow

> my preference would be to put it into this PR and then on our side we can do a single pull, test, and merge when everything is in place

Great. Then I'll keep pushing to this PR until it's ready to test/merge. Let's then also limit the scope of this PR to the bare minimum. We could get rid of the whole versioning discussion, if I just pin the versions to the ones used in current master. (We could then move that discussion to another issue / PR if you like). That would leave:

- [x] ship config templates with python package and provide cli command for writing the templates to disk
- [x] solve logging issues
- [x] ship the ui together with the python package and provide an easy way for using it

Does that sound reasonable? I'll have to see how much time I can invest in the next week or two, but if I can squeeze a litte bit of time out of my schedule it shouldn't take longer than that to get it done.

This is a really nice github interaction btw 😃 , looking forward to working on this with you!

-Andreas

choosehappy:
**versioning**


>  That option boils down to checking the version of scikit-image in the HistoQC code and then call remove_small_holes with the correct arguments for that version.

I was thinking about this, but given that the version .16 is already quite old (compared to .18), it doesn't make sense to me to carry on adding spaghetti code into the future

Better to make a hard break and version appropriately. 

That said, I think by your logic, since the latest version will break compatiability, isn't it more like:
1) current master tagged as v1.0.1
2) new release becomes v2.0.0, since the major version requires the non-compatible code change?


**logging**

> Is there any reason why the log can't be written directly to the output folder?

Yes, the logging is started before the output directory exists (line 17 versus line 191) : )  In fact, the logging begins before even the command line arguments are parsed, in case there is an error with the command line arguments which need to be logged


**ui**

sorry not sure i undestand the solution your proposing here, can you explain a bit more how you would see it work? is it cross o/s compatible?


**workflow**

Looks like a great way forward!


regarding this point:

> We could get rid of the whole versioning discussion, if I just pin the versions to the ones used in current master. 

It might be easier if we address this first before the others, inline with what i proposed above in versioning?

essentially do the bump now and move to the next major version which begins to have these other features?

If you like that, I can make the modification relatively immediately. Thoughts? 



> This is a really nice github interaction btw 😃 , looking forward to working on this with you!

Me as well, thanks for all your effort!

ap--:
#### versioning 

> That said, I think by your logic, since the latest version will break compatiability, isn't it more like:
> current master tagged as v1.0.1
> new release becomes v2.0.0, since the major version requires the non-compatible code change?

Yes, I think you're right. Just so I don't misunderstand what you wrote:
1. you add the changes that are required to support the newest scikit image and tag the newest version as v1.0.1
2. I rebase this PR on top of new master
3. once the PR is merged histoqc is tagged v2.0.0

or do you plan to bump the version to v2.0.0 once you introduce the scikit image update?

But to not get lost in the versioning discussion for too long: I would recommend to just do whatever you think confuses your primary user base the least. As long as we increase some part of the version number, it'll be clear that it's the newer version of the tool :)

#### ui

So when you use `python -m http.server` all it does is basically [spawn a `HTTPServer` instance]( https://github.com/python/cpython/blob/b4fc44bb2d209182390b4f9fdf074a46b0165a2f/Lib/http/server.py#L1256-L1293) that serves the contents of the current folder on `http://127.0.0.1/`

It's easy to change the behavior of that server dependent on the url path that the browser requests. So if you access `http://127.0.0.1/SomePath/Something` you have full control what is served to the browser if you customize the `HTTPServer` class. That makes it possible to serve the UI from some location on `http://127.0.0.1/` and "mount" an output directory to `http://127.0.0.1/Data` without having to change anything in the ui.

#### logging and pr

Great. I think I understand the logging requirements now and will make the necessary changes to make it work.

Feel free to make the changes for the version bump and tagged release at any time. I can easily rebase this PR when needed.

choosehappy:
> or do you plan to bump the version to v2.0.0 once you introduce the scikit image update?

Yes, this is what i was thinking. tag current, update scikit image, tag as 2.0, and then rebase to that

Thoughts?


**ui**

ah, i see where you're going with this. i took a look at their server code, longer than i expected (1.3k lines). seems doable, not sure how much effort it would take though. i don't have a  sense if it is the most elegant approach or too heavy-handed for what needs to be delivered...something to think about

ap--:
> Yes, this is what i was thinking. tag current, update scikit image, tag as 2.0, and then rebase to that

it depends on which version you want to release when this PR is merged. I would consider having histoqc available as a python module a "major version number increase" worthy addition.

```
current > v1.0
|
o Make changes to use newer skimage [TAG A]
|
o Make changes to be pip installable [TAG B]
```

So this way the first pip installable version would be `v3.0.0`.


There is of course the option to not consider the intermediate state a new release and include it in the changelog of the final state:

```
current > v1.0
|
o Make changes to use newer skimage
|
o Make changes to be pip installable [TAG]
```

which would lead to `v2.0.0`

But that decision is up to you, and I have no strong opinion either way 😃

choosehappy:
ah okay, i see what you mean

well we already have a v1.0 which is quite close to this version:

https://github.com/choosehappy/HistoQC/tags

so we could do a minor version tag with the current source as v1.1

i would then apply skimage, and make that version v2.0, so that it can sit
there for as long as needed to get the pip work in place, that seems like a
good milestone

I'll put this on my list and hopefully get to it today

ap--:
Hi @choosehappy 

I had a little bit of time and started work on the relevant items from the todo list.
So far I haven't touched the UI part yet, but will soon.

It's still very much work in progress, and I am currently building the bits and pieces to then basically replace most of the code in `__main__.py` (the former `qc_pipeline.py`) with a more compartmentalized and easier to test and reason about codebase.

But it's moving along nicely, and I'll keep you updated here on my progress.

Cheers,
Andreas 😃

choosehappy:
Wow, been a long time since I've seen enterprise-grade python code, looking
good! :)

I just did the tagging + dependency upgrading, should be good to go on that
front!

Contributing my little piece to this process : )

ap--:
I just rebased and force pushed the PR on current master.
I also added the ui to the package. It's accessible via running `python -m histoqc.ui`.
The logging issues should be fixed now.

But we need to do some testing, if it works as intended.

One open question would be if the default: `models`, `pen`, and `templates` data folders should be included in the module too.
I guess they probably should.

So the remaining things to be done:
- [x] do some testing
- [x] add cli commands to write config templates to disk
- [x] add cli commands to write UI directory structure to disk (so that existing workflows can be supported)
- [x] ? add default model, pen, templates resources to module

Cheers,
Andreas

ap--:
I just noticed that I didn't push the most recent commits.
Will update later today. Sorry if it caused confusion

choosehappy:
no worries : )

ap--:
I pushed the missing bugfixes 😃

ap--:
Hi @choosehappy,

This is now ready to be merged and I added a bunch of tests too.
Feel free to review it thoroughly.

Cheers,
Andreas 😃 

P.S.: I won't be able to spent time on this for quite a bit, but I'll be happy to answer questions that come up in the meantime.

choosehappy:
a super impressive effort, thanks so much! Lots of code to review might take me a bit, but I'll be sure to post questions here as they come up. Thanks again!

choosehappy:
Finally having a chance to sit down and play with this.

I'm encountering  an error I was hoping you could help with

In the HistoQC directory, I execute this:

```
python -m histoqc -n 4 "D:\research\IHC\CD20\*.svs"
```

and things start-up as expected, great:

```
2021-04-16 16:49:20,577 - WARNING - Configuration file not set (--config), using default
2021-04-16 16:49:20,604 - INFO - the pipeline will use these steps:
2021-04-16 16:49:20,605 - INFO -                BasicModule     getBasicStats
2021-04-16 16:49:20,613 - INFO -                ClassificationModule    byExampleWithFeatures:pen_markings
2021-04-16 16:49:20,613 - INFO -                ClassificationModule    byExampleWithFeatures:coverslip_edge
2021-04-16 16:49:20,614 - INFO -                LightDarkModule getIntensityThresholdPercent:tissue
2021-04-16 16:49:20,616 - INFO -                LightDarkModule getIntensityThresholdPercent:darktissue
2021-04-16 16:49:20,616 - INFO -                BubbleRegionByRegion    detectSmoothness
2021-04-16 16:49:20,619 - INFO -                MorphologyModule        removeFatlikeTissue
2021-04-16 16:49:20,620 - INFO -                MorphologyModule        fillSmallHoles
2021-04-16 16:49:20,620 - INFO -                MorphologyModule        removeSmallObjects
2021-04-16 16:49:20,621 - INFO -                BlurDetectionModule     identifyBlurryRegions
2021-04-16 16:49:20,621 - INFO -                BasicModule     finalProcessingSpur
```

When it processes the files, it error out here:

```
2021-04-16 16:49:22,943 - INFO - 0001_CD20_01.svs.svs -         ClassificationModule.byExample: pen_markings
2021-04-16 16:49:22,944 - INFO - 0001_CD20_01.svs.svs - Training model ClassificationModule.byExample:pen_markings
2021-04-16 16:49:22,949 - ERROR - D:\research\IHC\CD20\0001_CD20_01.svs.svs - Error analyzing file (skipping):   <class 'FileNotFoundError'> File not found. No such file: 'D:\research\histoqc_branch\D'
```

When tracing this error through, it is in ClassificationModule.py, and its cause is that the value in params["examples"] is incorrect

it contains:

```
'D:\\research\\histoqc_branch\\.histoqc_pkg_data_tmpq5_y3vbz\\./pen/1k_version/pen_green.png:D:\\research\\histoqc_branch\\.histoqc_pkg_data_tmpq5_y3vbz\\./pen/1k_version/pen_green_mask.png
```


which after going through the for loop results in:
```
['D',
 '\\research\\histoqc_branch\\.histoqc_pkg_data_tmpq5_y3vbz\\./pen/1k_version/pen_green.png',
 'D',
 '\\research\\histoqc_branch\\.histoqc_pkg_data_tmpq5_y3vbz\\./pen/1k_version/pen_green_mask.png']
```

which is clearly not going to work...it should contain the value (which was in the default config):

```
./pen/1k_version/pen_green.png:./pen/1k_version/pen_green_mask.png
```

which then gets modified into:

```
[./pen/1k_version/pen_green.png, ./pen/1k_version/pen_green_mask.png]
```

And that can correctly load the masks

It looks like the packaging is taking the relative directory from the config and adding on an incorrect absolute directory to it. 

Do you see what I mean? I suspect you know how to fix this much easier than i do : )

ap--:
Ah that's a good find. Interestingly I think it's a bug in the histoqc code that was now uncovered due to the packaging. If you would create a config.ini that contains paths for the classification module which contain ":" (for example absolute paths on windows) it will cause this bug.
That's because ":" is used as a list separator.

I'll think about a way to fix it with minimal changes and will push.
Probably later this week/early next week 😃 

Cheers,
Andreas

choosehappy:
Interesting point

HistoQC doesn't need to use ":", we can certainly use another seperator
which isn't available for paths, such as a vertical bar "|"

it maybe a good idea to switch to that to avoid any possible other issues?
thoughts?

ap--:
Sorry for the delay on my side, it's been a busy week here 😅 

For the sake of backwards compatibility I tried to fix it in a minimally invasive way.
As I see it, the problematic case occurs only on windows.

So now I only split on ":" that do not match "C:\\" etc...

choosehappy:
Quick update on this, just pulled this and merged in the latest changes from v2.1, did a full code review and tested it out. 

This is some *amazing* high-quality code, I can't thank @ap--  enough! I'll admit I had to go back to the python documentation to understand some of these changes : ) 

Lots of added quality of life components as well, for example the httpd server. magnificent!

Next step: I'll update the documentation and push to master, and tag as v3.0!

ap--:
Thank you 😊  
If it helps, I can rebase this PR on current master to make merging easier.

It's been a lot of fun working on this, and there's at least one more PR I can open once it's merged 😃

choosehappy:
i'll take care of the merge, gives me a chance to look through all the code and better understand it

A question I had,

post-installing histoqc as a module, is there a way to access the internal config files by name? or how would one access it?

When I'm in the histoqc directory, I access the config files in your version like this:

```
python -m histoqc -c histoqc\config\config_v2.1.ini -n 3 "*.svs"
```

but it seems like since the configs are within the package, there should be another command line option to use them, or am I misunderstanding this?

The only way I see would be use histoqc.config to extract one, save it in the same directory, and then use it, which I guess is similar to what we discussed, but I had thought that would only be required if one wants to modify the config

Thoughts?

choosehappy:
Just pushed everything.

If you wouldn't mind, can you take a look and see what you think?

in particular the readme.md and changelog

Additionally, I would be very happy to give you some well-deserved credit here, how can I appropriate attribute you some credit in the changelog?

ap--:
> but it seems like since the configs are within the package, there should be another command line option to use them, or am I misunderstanding this?

> The only way I see would be use histoqc.config to extract one, save it in the same directory, and then use it, which I guess is similar to what we discussed, but I had thought that would only be required if one wants to modify the config

Yes, it's currently designed like this. I assumed that users would always want to change the configurations, and so I thought of them more as example templates to get started with your own configuration. But it would be super easy to add another commandline option for selecting a predefined config that's stored in the package (basically it would make 'default' configurable here:)

https://github.com/choosehappy/HistoQC/blob/979d25b703af311f4b7ee63bd1cfb5ba3086606d/histoqc/__main__.py#L74-L79


> If you wouldn't mind, can you take a look and see what you think?

I had a quick look and spotted the following issues:
- because `dill` was added to `BaseImage.py`, the `install_requires` section in `setup.cfg` would need to be updated with the new dependency. I would suggest adding it without the version (the requirements.txt file already pins the version)
- I noticed in the readme, that all the cmdline --help sections print `__main__.py`. I remember fixing this exact issue in another package, I just need to find the code again to make argparse print the actual command it was called with
- in the readme you recommend installing via `python setup.py install`. This works and will be fine for almost all use cases. If you'd want to adhere to the PyPA recommendations change it to `pip install .` or the more robust variant `python -m pip install .` So to install from current master I would write:
  ```console
  $ git clone https://github.com/choosehappy/HistoQC.git
  $ cd HistoQC
  $ python -m pip install --upgrade pip  # (optional) upgrade pip to newest version
  $ pip install -r requirements.txt      # install pinned versions of packages
  $ pip install .
  ```

 I will have a more thorough look on the weekend. But the tests pass, so that is usually a good sign 😃 

> Additionally, I would be very happy to give you some well-deserved credit here, how can I appropriate attribute you some credit in the changelog?

Replacing my GitHub handle with my name and a link would be great. I guess to be overly correct the following should do the trick:
<pre>[Andreas Poehlmann](https://github.com/ap--) _(Machine Learning Research - Bayer AG)_</pre>

choosehappy:
Regarding the configuration: I agree almost all users should be modifying their configuration file to match their use case
From a "tool developer" standpoint, I think you essentially want to give folks something that they can run with a single line to get a flavor for the value of the tool and see what the outputs would be. 

As I'm typing that I realize that we essentially already have this with the default config (which I’m not in love with and is likely overkill for a 'first-user run' experience). 

Maybe instead I should modify the default to be something more generally useful. I’ll need to think about this, but if you have any thoughts would of course be open

Thanks for review, points well received and just pushed a set for addressing all of them!

Issue #192, opened 2021-07-01
UI: otsu and otsulocal masks aren't displayed because not in global_config.js


bswhite:
Masks created by getIntensityThresholdOtsu either with global thresholding (suffix _otsu.png) or local thresholding (suffix _otsulocal.png) don't display in the UI because they aren't listed in DEFAULT_IMAGE_EXTENSIONS in histoqc/ui/UserInterface/scripts/global_config.js.

Further, getIntensityThresholdOtsu hardcodes its suffix to "otsu" or "ostulocal." Other modules seem to instead leave this user-defined as in:
name = params.get("name", "classTask")

I suggest:
1. Removing name = "otsu" and name += "local" from getIntensityThresholdOtsu and replacing with name = params.get("name", "classTask")
Note that allowing a user-supplied name would allow us to display the corresponding mask in the UI by co-opting a name in global_config.js that is unused in the configuration (e.g., _bright might be used since the getIntensityThresholdPercent that outputs it would likely be replaced by getIntensityThresholdOtsu).
2. Adding _otsu.png and _otsulocal.png to global_config.js. This isn't technically necessary given (1), but will cover the common case where a user names their module 'otsu'. 

I'm happy to make such a pull request if @choosehappy agrees and would like me to.

choosehappy:
Thanks for your comments!

This Otsu method as you point out has not been well developed nor maintained. It was one of the earlier attempts at detecting tissue, but didn't appear to work very robustly across the full gamut of slides. The major challenge here being that tissue detection is the first step in the pipeline, and if that doesn't work, the rest of it doesn't work properly.

In the latest version of histoqc we've provided a new approach ( `LightDarkModule.minimumPixelIntensityNeighborhoodFiltering`) which has worked quite well for me across both IHC + H&E images, and has now become my "go to" approach.

That said, I am of course happy to accept a pull request for your suggested improvements, just want to make sure that the use case you're aiming to address wouldn't already be addressed by that new detection approach? 

If it is, and works well, we may even consider officially retiring the Otsu approach and removing it from the code base

Thoughts?

bswhite:
Thanks for the background on Otsu and your experience with it.

[...]

PR #193, opened 2021-07-07
Ensure Otsu displays in UI

bswhite:
- Added 'otsu' and 'otsulocal' as file extensions in global_config.js that will be displayed in UI.
- Allowed name of Otsu method to be configurable in LightDarakModule.py (i.e., not restricted to otsu or otsulocal). Note that local is not automatically appended if local thresholding is performed.


choosehappy:
Looks great, thanks! relates to #192

PR #196, opened 2021-07-22
use multi-stage build in Dockerfile to halve size

kaczmarj:
This commit proposes using a multi-stage Dockerfile. In the first stage,
we use the official python:3.8 image, which includes a C compilers. We
install the project and its dependencies. We create a dedicated virtual
environment for the project. Then in the second stage, we copy over the
project files and the python virtual environment, and we install
libopenslide0, which is required for openslide-python runtime.

This results in a more complex Dockerfile, but the resulting image is
half as big as the original (671 MB from about 1.3GB). I am also happy
to maintain this Dockerfile and answer any questions about it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ COPY . .
 # installation into the second stage of the build.
 ENV PATH="/opt/HistoQC/venv/bin:$PATH"
 RUN python -m venv venv \
+    && python -m pip install --no-cache-dir setuptools wheel \
     && python -m pip install --no-cache-dir -r requirements.txt \
     && python -m pip install --no-cache-dir . \
     # We force this so there is no error even if the dll does not exist.
@@ -20,13 +21,13 @@ RUN python -m venv venv \
 
 FROM python:3.8-slim
 ARG DEBIAN_FRONTEND=noninteractive
-WORKDIR /opt/HistoQC
-COPY --from=builder /opt/HistoQC/ .
-ENV PATH="/opt/HistoQC/venv/bin:$PATH"
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         libopenslide0 \
         libtk8.6 \
     && rm -rf /var/lib/apt/lists/*
+WORKDIR /opt/HistoQC
+COPY --from=builder /opt/HistoQC/ .
+ENV PATH="/opt/HistoQC/venv/bin:$PATH"
 
 WORKDIR /data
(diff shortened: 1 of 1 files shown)


kaczmarj:
## build docker image

in the root directory (where the Dockerfile is), run the following

```bash
docker build --tag histoqc .
```

## run histoqc in docker

we mount things like `/etc/passwd` and `/etc/group` so that we can run the docker image as the current user. if we don't do this, then the files created in the container will be owned by root.

```bash
docker run --rm -it \
  -v /etc/passwd:/etc/passwd:ro \
  -v /etc/group:/etc/group:ro \
  --user $(id -u):$(id -g) \
  -v /path/to/wsi:/data kaczmarj/histoqc \
    python -m histoqc --outdir /data/histoqc /data/slide.svs
```

the outputs are in `/path/to/wsi/histoqc`

choosehappy:
Apologies for the delay, finally had a chance to look into this.

Some statistics on my machine from the current Dockerfile
```
Previous Docker File
[+] Building 133.2s (8/8) FINISHED
REPOSITORY   TAG       IMAGE ID       CREATED          SIZE
histoqc      latest    9a4497b29779   14 seconds ago   1.15GB
```

And from your Dockerfile
```
New Docker file
[+] Building 99.7s (15/15) FINISHED
REPOSITORY   TAG       IMAGE ID       CREATED          SIZE
histoqc      latest    bc1e4b313941   About a minute ago   737MB
```

So overall the build time is 25% faster, and 35% smaller, both very nice properties : ) 


I wanted to give a moment for @petrovm3 to chime in, if they have any comments, since they've done a lot of work on the docker building previously.

Perhaps @ap-- has thoughts as well?

ap--:
That's great 👍 
tbh, I've been using a micromamba based docker container for now, and I just checked and it's ~2.7GB 😅 So I should definitely switch.

kaczmarj:
thanks! i built the original docker image and the one in this pr. here are my results

```
$ docker images
REPOSITORY   TAG        IMAGE ID       CREATED              SIZE                                                                                                                             
histoqc      pr         45fc57f2a87d   8 seconds ago        690MB                                                                                                                            
histoqc      original   f259ed95eaf7   About a minute ago   1.15GB
```

build times: 96 s for original, 52 s for this pr. there is a small discrepancy in final image size from https://github.com/choosehappy/HistoQC/pull/196#issuecomment-900297022 because of my latest commit in this pr.

choosehappy:
looks good to me, thanks for your efforts!

PR #198, opened 2021-08-20
add expanduser in basepath and outdir

### LINKED ISSUE #197: using `~` in outdir makes a directory named `~`
When I specify an out directory with `~`, as in the example below, a directory named `~` is created in my current working directory. I don't know where the fix should go, but it probably involves adding [`os.path.expanduser`](https://docs.python.org/3/library/os.path.html#os.path.expanduser) somewhere.

```bash
python -m histoqc \
  --config config_first.ini \
  --basepath ~/data/wsi/ \
  --outdir ~/data/wsi-artifacts/qc-inital \
  '*'
```

kaczmarj:
Closes https://github.com/choosehappy/HistoQC/issues/197

When using a path with a tilde `~` on unix-like systems, the path is
interpreted as a literal `./~`, meaning a directory named `~` in the
current working directory. Users typically do not want this and would
probably be surprised by this behavior. I know I was surprised when I
ran histoqc and then saw a new directory named `~`. I removed it very
carefully, trying to be sure I didn't trash my home directory :)

This commit fixes the issue by expanding the tilde into the path to the
home directory. If a path does not begin with a tilde, the path is
unchanged.

### DIFF
--- a/histoqc/__main__.py
+++ b/histoqc/__main__.py
@@ -95,7 +95,7 @@ def main(argv=None):
             return -1
 
     # --- create output directory and move log --------------------------------
-
+    args.outdir = os.path.expanduser(args.outdir)
     os.makedirs(args.outdir, exist_ok=True)
     move_logging_file_handler(logging.getLogger(), args.outdir)
 
@@ -119,7 +119,7 @@ def main(argv=None):
     results.add_header(f"command_line_args:\t{' '.join(argv)}")
 
     # --- receive input file list (there are 3 options) -----------------------
-
+    args.basepath = os.path.expanduser(args.basepath)
     if len(args.input_pattern) > 1:
         # more than one input_pattern is interpreted as a list of files
         # (basepath is ignored)


choosehappy:
good catch! our home directories on our machines all have quotas so we tend to not to write experimental results to them, otherwise this issue would have popped up.

I'm pretty surprised, I thought this tilde would be expanded by bash before sending, but it seems it isn't the case. thanks for the commit!

kaczmarj:
i was surprised too! perhaps it's because i was in a `screen` session... not sure. thanks for merging.

choosehappy:
i did some reading before accepting the merge : ) non-expansion looks to be the expected results in a few cases, more full discussion in the reference manual:

https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html
https://askubuntu.com/questions/890415/why-couldnt-i-use-instead-of-home-username-when-giving-the-file-pat

regardless, we should have your code which looks for and compensates for it

BTW, I switched to TMUX instead of screen a number of years ago, and haven't looked back. if you haven't heard of it, I would check it out!

kaczmarj:
excellent, thanks for the references. i definitely should switch to TMUX. just need to learn the keyboard shortcuts :)

choosehappy:
i hear you, its a bit of a hassle, but I can assure you, in this case, its definitely worth it : )

PR #200, opened 2021-08-21
remove Python::3.6 classifier

kaczmarj:
This removes `Programming Language :: Python :: 3.6` from the classifiers, because according to `python_requires`, histoqc requires at least python 3.7.


choosehappy:
Thanks. I see why this happened. We  bumped up the scikit-image version, which comes with this caveat:

_This release of scikit-image drops support for Python 3.6 in accordance with the NEP-29 Python and Numpy version support community standard: Python 3.7 or newer is required to run this version._

Issue #201, opened 2021-09-08
Configuration Changes needed


tasvora:
In order to run Histoqc , I had to add the full path for the config file
path to the config file
python -m histoqc -c histoqc/config/config_v2.1.ini -n 3 "*.svs"

Need to resolve this.

choosehappy:
I think the logic should be

if args.config file exists --- > config.read (this is currently done)
if args.config file does not exist, try to load a template with that name
(this is not done)
if args.config does not exist, load the template default (this is currently
done)

PR #203, opened 2021-09-20
Pull Request for Configuration Changes #201

tasvora:


histoqc/config/__init__.py
```diff
@@ -34,5 +34,8 @@ def read_config_template(name=None):
     """return the contents of a configuration template"""
     templates = list_config_templates()
     if name not in templates:
-        raise KeyError(f'no configuration template found under key {name!r}')
```
choosehappy: I think you're working too hard here, i suspect its going to be a lot easier and more robust to instead modify CONFIG_TEMPLATE_RE to accept all desired config files, instead of hacking around its failure at this stage

histoqc/__main__.py
```diff
@@ -76,7 +76,12 @@ def main(argv=None):
         lm.logger.warning(f"Configuration file not set (--config), using default")
         config.read_string(read_config_template('default'))
     else:
-        config.read(args.config)
+        try:
```
choosehappy: I think you want an elif  pattern here, with an explicit check to see if the file "args.config" exists

some general explanation:
https://www.pythonforbeginners.com/control-flow-2/when-to-use-try-catch-instead-of-if-else
https://www.geeksforgeeks.org/try-except-vs-if-in-python/
https://stackoverflow.com/questions/1835756/using-try-vs-if-in-python

so one should ask how frequently the user will take this approach? and if it is frequent, an "if" is more appropriate (and faster)

note as well that this "except" creates a slightly unexpected situation. what if the file *does* exist, but it is not parseable? i.e., you're setup to catch one error but not another

choosehappy (changes requested):
Its getting there, made a few comments

histoqc/config/__init__.py
tasvora: The regular expressions obtain the template name, for example config_clinical.ini, template name is clinical.
For config_v2.1.ini , template name is v2.1
Will have change in command - python -m histoqc -c v2.1.ini -n 3 "*.svs"
Is that fine? The current fix lets it to be used as is

choosehappy: Interesting! I hadn't noticed or considered that....

That said, it doesn't seem to work:
```
D:\temp\HistoQC>python -m histoqc -c v2.1.ini "D:\research\cetl_vs_mct\B15_6130_2_CETL.svs"
2021-09-23 11:29:44,558 - WARNING - Configuration file v2.1.ini not found, checking for a template match.
Traceback (most recent call last):
  File "D:\temp\HistoQC\histoqc\__main__.py", line 80, in main
    config.read_file(open(args.config))
FileNotFoundError: [Errno 2] No such file or directory: 'v2.1.ini'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\temp\HistoQC\histoqc\__main__.py", line 229, in <module>
    sys.exit(main())
  File "C:\Python37\lib\contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "D:\temp\HistoQC\histoqc\__main__.py", line 83, in main
    config.read_string(read_config_template(args.config))
  File "D:\temp\HistoQC\histoqc\config\__init__.py", line 40, in read_config_template
    raise KeyError(f'no configuration template found under key {name!r}')
KeyError: "no configuration template found under key 'v2.1.ini'"

```

but I think based on what you said, the name should actually be "v2.1" and not "v.2.1.ini" (minus file extension)?

Regardless that also doesn't work unfortunately

```
D:\temp\HistoQC>python -m histoqc -c v2.1 "D:\research\cetl_vs_mct\B15_6130_2_CETL.svs"
2021-09-23 11:29:36,371 - WARNING - Configuration file v2.1 not found, checking for a template match.
Traceback (most recent call last):
  File "D:\temp\HistoQC\histoqc\__main__.py", line 80, in main
    config.read_file(open(args.config))
FileNotFoundError: [Errno 2] No such file or directory: 'v2.1'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\temp\HistoQC\histoqc\__main__.py", line 229, in <module>
    sys.exit(main())
  File "C:\Python37\lib\contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "D:\temp\HistoQC\histoqc\__main__.py", line 83, in main
    config.read_string(read_config_template(args.config))
  File "D:\temp\HistoQC\histoqc\config\__init__.py", line 40, in read_config_template
    raise KeyError(f'no configuration template found under key {name!r}')
KeyError: "no configuration template found under key 'v2.1'"


```


But this is a highly related question which I think also needs to be addressed and doing so would address the issue above:
why doesn't the config_v2.1.ini appear here, if it is in the directory?

```
D:\research\HistoQC>python -m histoqc.config --list
# available example configurations:
- default:  config.ini
- clinical: config_clinical.ini
- first:    config_first.ini
- ihc:      config_ihc.ini
- light:    config_light.ini
```

tasvora: Currently the regular expression does not account for decimal points, and will have to add it in,
Once added it will pick the name v2.1 and it will appear as a template name.

Should I go ahead and change it ?

choosehappy: yes, i think so, thanks

tasvora: now we are adding a decimal point as well in the CONFIG_TEMPLATE_RE
CONFIG_TEMPLATE_RE = re.compile(r'config(_(?P<name>[A-Za-z][A-Za-z0-9.]+))?[.]ini')
```
>python -m histoqc.config --list**

- default:  config.ini
- clinical: config_clinical.ini
- first:    config_first.ini
- ihc:      config_ihc.ini
- light:    config_light.ini
- v2.1:     config_v2.1.ini
```
Hence now the user can use the command 
```
> python -m histoqc -c v2.1 -n 3 "*.svs" and it will work fine.
```

histoqc/__main__.py
tasvora: may be I did not think correctly.
I am assuming for a user to use -c config parameter, the input type will be one of the listed configurations.
```
- default:  config.ini
- clinical: config_clinical.ini
- first:    config_first.ini
- ihc:      config_ihc.ini
- light:    config_light.ini
- v2.1:     config_v2.1.ini
```
Example
```
> python -m histoqc -c v2.1 -n 3 "*.svs"
or
> python -m histoqc -c light -n 3 "*.svs"
```

Is there a possibility that the user will have a filepath given ?
Current code is incorrect, hence it will not work at all if any config parameter is provided, and to make it work 
```
config.read(args.config) line 79
needs to be changed to
 config.read_string(read_config_template(args.config))
```

tasvora: try catch will be removed completely

tasvora: Or we can just change it to 
```
    if args.config:
        if args.config.endswith('.ini'):
            config.read(args.config) # if file name with path provided
        else:
            config.read_string(read_config_template(args.config)) #template name provided
    else:
        lm.logger.warning(f"Configuration file not set (--config), using default")
        config.read_string(read_config_template('default'))
```

histoqc/config/__init__.py
choosehappy: Love it!

histoqc/__main__.py
choosehappy: I think we're looking for something like this:

```
if not args.config: #not set, load default
    lm.logger.warning(f"Configuration file not set (--config), using default")
    config.read_string(read_config_template('default'))
elif os.path.exists(args.config): #check to see if exists as a file with higher priority
    config.read_file(open(args.config))
else: #check to see if named template
    lm.logger.warning(f"Configuration file {args.config} assuming to be a template...checking.")
    config.read_string(read_config_template(args.config))
    
```

tasvora: Ah okay, no problem will change it and send a pull request 
again

choosehappy: Please make sure to test it first, I wrote that code in notepad : )

tasvora: Yes I will

choosehappy (approved):

PR #221, opened 2022-09-13
Unified interface in BaseImage for images other than openslide-supported format and more support for usability in the python scripts or interactive shells

CielAl:
1. Add an ImageHandle wrapper class for openslide or PIL-supported data, all image data-related operations in the BaseImage access the interface of the ImageHandle derivated object first.  The naming of methods in the interface mostly follows the style of openslide APIs for the convenience of coding. 
2. For easier access and usage of histoqc APIs, PIL-supported data can be created by either a valid URI pointing to an image, or from an existing ndarray or PIL.Image.Image object.
3. Functionality/APIs/Modules to extract tiles in the obtained img_mask_use either in the CLI pipeline or on-the-fly in an existing python script workflow. A thumbnail with tile bounding boxes is rendered to identify the location of extracted tiles.
4. Some potential bug fixing:

- in several modules, certain boolean parameters are loaded from config using strtobool(params.get(key, True)), with a boolean type default value. However, strtobool can only accept str type as the inputs. Hence, reading from default values will crash the pipeline.
- in https://github.com/choosehappy/HistoQC/blob/master/histoqc/BubbleRegionByRegion.py#L61, somehow the level parameter is not used and the code always read regions from level 1 -- this seems to be a typo but need further confirmation.
- According to https://github.com/openslide/openslide-python/issues/115, in python3.8+, the behavior of loading dlls is changed. Thus, the previous trick to add path info no longer works to import openslide. Hence, a unified wrapper of openslide importing is added into histoqc._import_openslide.  Openslide can be imported from the wrapper in all project files.

5. Add and pass the pytest files for the corresponding changes.
sample


CielAl:
image

PR #225, opened 2022-10-12
Add ability to use a server-side results file

### LINKED ISSUE #224: "View Results" button in UI creates problems when using port forwarding
I'm running the core HistoQC tool in a large VM in the cloud. When I now want to view the results with `histoqc.ui`, I'm in a bit of trouble:  I can't run the UI directly on the VM, so I need to use port forwarding to my laptop. When I now click the "View results" button, it would open a file dialog on my laptop - but of course my results still live in my cloud VM.

I can copy the `results.tsv` file to my laptop, but that of course makes the workflow more cumbersome.

Do you have any suggestions how we can better proceed? How could we add a way of bypassing that "View Results" button and provide a result file on the commandline, for example? This would cover a lot of our `histoQC` usecases already.

@harshita-s FYI

ant0nsc:
Closes #224

### DIFF
--- a/histoqc/ui/UserInterface/index.html
+++ b/histoqc/ui/UserInterface/index.html
@@ -61,9 +61,13 @@
 	<div id="upload-button" class="container pt-md-5 text-center">
 		<label>
 			<span class="btn btn-primary btn-lg">
-				View Results <input id="upload-input" type="file" accept=".tsv" multiple>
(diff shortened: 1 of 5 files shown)


jacksonjacobs1:
HistoQC's v2 UI currently requires the results.tsv file to be server-side.

More specifically, all results.tsv files must reside in the same directory containing output masks.

PR #226, opened 2022-11-18
bounding_box and best level for downsample

JulienMassonnet:
Merge alpha channel to avoid black background.
Add bounding_box option on config file to open WSI on the bounding box (default to False).
Find best level for downsample when factor isn't integer.

### DIFF
--- a/histoqc/BaseImage.py
+++ b/histoqc/BaseImage.py
@@ -4,6 +4,7 @@ import numpy as np
 import inspect
 import zlib, dill
 from distutils.util import strtobool
+from PIL import Image
 
 #os.environ['PATH'] = 'C:\\research\\openslide\\bin' + ';' + os.environ['PATH'] #can either specify openslide bin path in PATH, or add it dynamically
 import openslide
@@ -61,7 +62,16 @@ class BaseImage(dict):
         self["dir"] = os.path.dirname(fname)
 
         self["os_handle"] = openslide.OpenSlide(fname)
-        self["image_base_size"] = self["os_handle"].dimensions
+        self["bounding_box"] = strtobool(params.get("bounding_box",'False'))
+        if self["bounding_box"] == True:
+            try:
+                self["image_base_size"] = (int(self["os_handle"].properties.get(openslide.PROPERTY_NAME_BOUNDS_WIDTH,'NA')),
+                                           int(self["os_handle"].properties.get(openslide.PROPERTY_NAME_BOUNDS_HEIGHT,'NA')))
+            except:
+                logging.error(f"{self['filename']}: could not read the size of the bounding box")
+                exit()
+        else:
+            self["image_base_size"] = self["os_handle"].dimensions
         self["image_work_size"] = params.get("image_work_size", "1.25x")
         self["mask_statistics"] = params.get("mask_statistics", "relative2mask")
         self["base_mag"] = getMag(self, params)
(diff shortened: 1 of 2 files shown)


histoqc/BaseImage.py
```diff
@@ -61,7 +62,16 @@ def __init__(self, fname, fname_outdir, params):
         self["dir"] = os.path.dirname(fname)
 
         self["os_handle"] = openslide.OpenSlide(fname)
-        self["image_base_size"] = self["os_handle"].dimensions
+        self["bounding_box"] = strtobool(params.get("bounding_box",'False'))
+        if self["bounding_box"] == True:
+            try:
+                self["image_base_size"] = (int(self["os_handle"].properties.get(openslide.PROPERTY_NAME_BOUNDS_WIDTH,'NA')),
+                                           int(self["os_handle"].properties.get(openslide.PROPERTY_NAME_BOUNDS_HEIGHT,'NA')))
+            except:
```
choosehappy, line 70: i don't think you want to treat this so aggressively if it fails

keep in mind you may be processing 10,000 WSI, and if _one_ of them fails, do you want the entire pipeline to stop? probably not

instead, its better to mark a warning in the log, and in the file, like this:

https://github.com/choosehappy/HistoQC/blob/979d25b703af311f4b7ee63bd1cfb5ba3086606d/histoqc/AnnotationModule.py#L74

and then to simply carry-on as if bounding_box == False (i.e., default previous behavior)

histoqc/BaseImage.py
```diff
@@ -98,11 +108,30 @@ def getImgThumb(self, dim):
         key = "img_" + str(dim)
         if key not in self:
             osh = self["os_handle"]
+            xc=yc=0
+            if self["bounding_box"] == True:
+                try:
+                    xc=int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_X,'NA'))
+                    yc=int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_Y,'NA'))
+                except:
+                    logging.error(f"{self['filename']}: could not find bounding box coordinates")
```
choosehappy, line 117: same comment as above

histoqc/BaseImage.py
```diff
@@ -98,11 +108,30 @@ def getImgThumb(self, dim):
...
                 dim = float(dim)
                 if dim < 1 and not dim.is_integer():  # specifying a downscale factor from base
                     new_dim = np.asarray(osh.dimensions) * dim
-                    self[key] = np.array(osh.get_thumbnail(new_dim))
+                    if self["bounding_box"] == True:
+                        downsample = max(*(dim / thumb for dim, thumb in zip(osh.dimensions, new_dim)))
+                        level = osh.get_best_level_for_downsample(downsample)
+                        size=tuple((np.array(self["image_base_size"])/osh.level_downsamples[level]).astype(int))
+                        tile = osh.read_region((xc, yc), level, size)
+                        bg_color = '#' + osh.properties.get(openslide.PROPERTY_NAME_BACKGROUND_COLOR, 'ffffff')
```
choosehappy, line 128: it looks like you do this multiple times through out your code

i would instead suggest making a function for the background stuff, and having that wrap around the call to read_region

histoqc/BaseImage.py
```diff
@@ -98,11 +108,30 @@ def getImgThumb(self, dim):
...
+                    yc=int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_Y,'NA'))
+                except:
+                    logging.error(f"{self['filename']}: could not find bounding box coordinates")
+                    return -1
             if dim.replace(".", "0", 1).isdigit(): #check to see if dim is a number
                 dim = float(dim)
                 if dim < 1 and not dim.is_integer():  # specifying a downscale factor from base
                     new_dim = np.asarray(osh.dimensions) * dim
-                    self[key] = np.array(osh.get_thumbnail(new_dim))
+                    if self["bounding_box"] == True:
```
choosehappy, line 123: usually in python folks would just say:

```
if self["bounding_box"]:
```

the == True part isn't needed

histoqc/BaseImage.py
```diff
@@ -98,11 +108,30 @@ def getImgThumb(self, dim):
...
+                        downsample = max(*(dim / thumb for dim, thumb in zip(osh.dimensions, new_dim)))
+                        level = osh.get_best_level_for_downsample(downsample)
```
choosehappy, line 125: shouldn't this get_best_level_for_downsample be addressed the same way as the corrected version you produced below? likely want to turn that into a function that wraps around the get_best_level_for_downsample  function in all places?

histoqc/BaseImage.py
```diff
@@ -112,14 +141,36 @@ def getImgThumb(self, dim):
...
-                    self[key] = np.array(osh.get_thumbnail((dim, dim)))
+                    if self["bounding_box"] == True:
+                        downsample = max(*(dim / thumb for dim, thumb in zip(osh.dimensions, (dim, dim))))
+                        level = osh.get_best_level_for_downsample(downsample)
+                        size=tuple((np.array(self["image_base_size"])/osh.level_downsamples[level]).astype(int))
+                        tile = osh.read_region((xc, yc), level, size)
+                        bg_color = '#' + osh.properties.get(openslide.PROPERTY_NAME_BACKGROUND_COLOR, 'ffffff')
+                        thumb = Image.new('RGB', tile.size, bg_color)
+                        thumb.paste(tile, None, tile)
+                        thumb.thumbnail(size, getattr(Image, 'Resampling', Image).LANCZOS)
```
choosehappy, line 170: here as well, the get_best_level_for_downsample  is likely not ideal, we can do better with the modifications you made on line 188

histoqc/BaseImage.py
```diff
@@ -133,23 +184,43 @@ def getImgThumb(self, dim):
                 target_mag = float(dim.upper().split("X")[0])
 
                 down_factor = base_mag / target_mag
-                level = osh.get_best_level_for_downsample(down_factor)
+                relative_down_factors_idx=[np.isclose(i/down_factor,1,atol=.01) for i in osh.level_downsamples]
```
choosehappy, line 187: this is the stuff that can be put into a function which gets called/wrapped

choosehappy (commented):
overall code looks very good, thanks! 

made some suggestions regarding refactoring so that you have a lot less duplicated code - > this code should go into a function and be reused everywhere instead of copy->pasted

can you take a look please?

Issue #219, opened 2022-09-08
Config files and LightDarkModule.getIntensityThresholdPercent


guillermojp:
Hi! Thanks a lot for the tool, it's very useful.

I'm paraphrasing a bit from another issue (#213).

[Line 61](https://github.com/choosehappy/HistoQC/blob/master/histoqc/LightDarkModule.py#L61) in the 'getIntensityThresholdPercent(s, params)' function in the 'LightDarkModule.py' file expects the variable name to be called "lower_variance". However, in the config files, the variable is often referenced as "lower_var", which causes the function to take the default value "-inf":

[config.ini](https://github.com/choosehappy/HistoQC/blob/master/histoqc/config/config.ini#L100)
[config_first.ini](https://github.com/choosehappy/HistoQC/blob/master/histoqc/config/config_first.ini#L34)
[config_light.ini](https://github.com/choosehappy/HistoQC/blob/master/histoqc/config/config_light.ini#L98)
[config_v2.1.ini](https://github.com/choosehappy/HistoQC/blob/master/histoqc/config/config_v2.1.ini#L112)
[config_clinical](https://github.com/choosehappy/HistoQC/blob/master/histoqc/config/config_clinical.ini#L100)

Thanks

choosehappy:
you are totally right, i see now with #213 they agreed to submit a pull
request but it looks like that never came to fruition :-\

would you be interested in submitting a pull-request to address this issue
so we can close em both?

guillermojp:
Sure, but... Do I change the names of the variables in the config files or the name of the variable read in the module? Either way some users will experience bugs.

choosehappy:
I would change the ones in the module, since that is the "single point" of
error? what do you think?

Issue #213, opened 2022-03-06
LightDarkModule.getIntensityThresholdPercent


mrigank-srivastava:
Hi!

Line number 65 in the 'getIntensityThresholdPercent(s, params)' function in the 'LightDarkModule.py' file is computing the standard deviation for the 'img' object, however, the variable name 'img_var' seems to want the variance which is also used in further comparisons within the function. All config files in which the above module is used specify the variance (upper and/or lower). Below is the code snippet from the function.

```python
    lower_thresh = float(params.get("lower_threshold", -float("inf")))
    upper_thresh = float(params.get("upper_threshold", float("inf")))

    lower_var = float(params.get("lower_variance", -float("inf")))
    upper_var = float(params.get("upper_variance", float("inf")))

    img = s.getImgThumb(s["image_work_size"])
    img_var = img.std(axis=2) ## this one

    map_var = np.bitwise_and(img_var > lower_var, img_var < upper_var)

    img = color.rgb2gray(img)
    map = np.bitwise_and(img > lower_thresh, img < upper_thresh)

    map = np.bitwise_and(map, map_var)

    s["img_mask_" + name] = map > 0
```

Can you provide any explanation for comparing standard deviation with the variance given in the config files?
Thanks!!

choosehappy:
OOops. thanks for spotting that, it looks like the variable name is
incorrect. For consistency with previous results, everything should be
based on standard deviation

would you be willing to submit a pull request to address this issue? : )

mrigank-srivastava:
> would you be willing to submit a pull request to address this issue?

Yes

PR #228, opened 2023-01-18
Try to fix the issue #224

### LINKED ISSUE #224: "View Results" button in UI creates problems when using port forwarding

nanli-emory:
add a result parameter to indicate the result file
add DEFAULT_RESULT_FILE_NAME if the user provided
automatically load the result if the **DEFAULT_RESULT_FILE_NAME** exited in the global_config.js

This pull request is trying to fix:
#224 -> [66fc7fa](https://github.com/choosehappy/HistoQC/pull/228/commits/66fc7faf861f8641c883fc7a43ba5034c53ebca2), [f53f274](https://github.com/choosehappy/HistoQC/pull/228/commits/f53f274f4039b7f26d35f70899618b224eb9ef1c)

#213 and #219 -> [45133c7](https://github.com/choosehappy/HistoQC/pull/228/commits/45133c71f8bc12b7c5df03c958c4a496f1fd7f30), [70a7613](https://github.com/choosehappy/HistoQC/pull/228/commits/70a76131021aecafa4527747f978ee72e2ff27d5)

### DIFF
--- a/histoqc/LightDarkModule.py
+++ b/histoqc/LightDarkModule.py
@@ -55,24 +55,14 @@ def getIntensityThresholdPercent(s, params):
     name = params.get("name", "classTask")
     logging.info(f"{s['filename']} - \tLightDarkModule.getIntensityThresholdPercent:\t {name}")
 
-    lower_thresh = float(params.get("lower_threshold", -float("inf")))
-    upper_thresh = float(params.get("upper_threshold", float("inf")))
+    lower_thresh = float(params.get("lower_threshold", "-inf"))
+    upper_thresh = float(params.get("upper_threshold", "inf"))
 
     # Prepare parameter names due to issues #213 and #219
     # set lower standard deviation
-    if params.get('lower_var'):
-        lower_std = float(params.get('lower_var'))
-    elif params.get('lower_std'):
-        lower_std = float(params.get('lower_std'))
-    else:
-        lower_std = float("inf")
+    lower_std = float(args.get("lower_var") or args.get("lower_std") or "-inf")
     # set upper standard deviation
-    if params.get('upper_var'):
-        upper_std = float(params.get('upper_var'))
-    elif params.get('upper_std'):
-        upper_std = float(params.get('upper_std'))
-    else:
-        upper_std = float("inf")
+    upper_std = float(args.get("upper_var") or args.get("upper_std") or "inf")
 
     img = s.getImgThumb(s["image_work_size"])
     img_std = img.std(axis=2)
--- a/histoqc/LightDarkModule.py
+++ b/histoqc/LightDarkModule.py
@@ -59,25 +59,10 @@ def getIntensityThresholdPercent(s, params):
     upper_thresh = float(params.get("upper_threshold", "inf"))
 
     # Prepare parameter names due to issues #213 and #219
-    
     # set lower standard deviation
-    lower_var = params.get("lower_var")
-    lower_std = float(lower_var or params.get("lower_std") or "-inf")
-    if lower_var:
-        logging.warning(f"{s['filename']} - LightDarkModule.getIntensityThresholdPercent:{name} "
-                        f"lower_var will be deprecated in future versions in favor of lower_std")
-        s["warnings"].append(f"LightDarkModule.getIntensityThresholdPercent:{name} "
-                        f"lower_var will be deprecated in future versions in favor of lower_std")
-
+    lower_std = float(args.get("lower_var") or args.get("lower_std") or "-inf")
     # set upper standard deviation
-    upper_var = params.get("upper_var")
-    if upper_var:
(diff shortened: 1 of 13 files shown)


histoqc/data/__init__.py
```diff
@@ -145,7 +145,12 @@ def _traverse_copy(traversable, root):
 
         if traversable.is_file():
             pth.write_bytes(traversable.read_bytes())
-
+            
+            # add default result file name as global variable
+            if result and name == 'global_config.js':
+                with open(pth, 'a') as f:
+                    f.write(f"// Default result file name \nvar DEFAULT_RESULT_FILE_NAME = \"{result}\";")
+
```
choosehappy, line 153: yea! nice hook!

histoqc/ui/__main__.py
```diff
@@ -26,6 +26,10 @@ def main(argv=None):
     parser.add_argument('--deploy', metavar="OUT_DIR",
                         default=None,
                         help='Write UI to OUT_DIR')
+    parser.add_argument('--result', '-rs', 
+                        type=str,
+                        default=None,
+                        help='If provided the result file name, use this fixed result file name rather than what is selected in the UI.')
```
choosehappy: can we add 1/2 a sentence to the help message ".,..., especially useful for remote data viewing"

choosehappy: can we add the word local as well - "use this fixed local result file"

choosehappy (approved):
looks good, 1 minor comment

histoqc/LightDarkModule.py
```diff
@@ -58,18 +58,31 @@ def getIntensityThresholdPercent(s, params):
     lower_thresh = float(params.get("lower_threshold", -float("inf")))
     upper_thresh = float(params.get("upper_threshold", float("inf")))
 
-    lower_var = float(params.get("lower_variance", -float("inf")))
-    upper_var = float(params.get("upper_variance", float("inf")))
+    # Prepare parameter names due to issues #213 and #219
+    # set lower standard deviation
+    if params.get('lower_var'):
```
choosehappy: yup, looks good and is really the only way to ensure backward compatibility due to my bug. 

can we add a warning to the log that lower_var will be deprecated in future versions in favor of lower_std?

we can then justify cleaning up the code in the future

histoqc/LightDarkModule.py
```diff
@@ -55,24 +55,14 @@ def getIntensityThresholdPercent(s, params):
...
 
     # Prepare parameter names due to issues #213 and #219
     # set lower standard deviation
-    if params.get('lower_var'):
-        lower_std = float(params.get('lower_var'))
-    elif params.get('lower_std'):
-        lower_std = float(params.get('lower_std'))
-    else:
-        lower_std = float("inf")
+    lower_std = float(args.get("lower_var") or args.get("lower_std") or "-inf")
```
choosehappy, line 63: wOoOo! super elegant. i had *no* idea that you could this in python (i actually fired up my interpreter to play with it). i had assumed it would turn simply a true/false value after evaluating the condition, but it actually returns the first non 0/None value? amazing!

histoqc/LightDarkModule.py
```diff
@@ -59,10 +59,25 @@ def getIntensityThresholdPercent(s, params):
...
     # set upper standard deviation
-    upper_std = float(args.get("upper_var") or args.get("upper_std") or "inf")
+    upper_var = params.get("upper_var")
+    if upper_var:
+        logging.warning(f"{s['filename']} - LightDarkModule.getIntensityThresholdPercent:{name} "
+                        f"upper_var will be deprecated in future versions in favor of upper_std")
+        s["warnings"].append(f"LightDarkModule.getIntensityThresholdPercent:{name} "
+                        f"upper_var will be deprecated in future versions in favor of upper_std")
+    
+    upper_std = float(upper_var or params.get("upper_std") or "inf")
```
choosehappy: sorry sorry please don't hate me...but i loved your previous 1-liner code better : ) 

I didn't realize that it was possible to set the statement in a single line like:

```
 upper_std = float(args.get("upper_var") or args.get("upper_std") or "inf")
```

which is clearly more elegant to me than having this deprecation notice (which i know i asked for...my fault)

Can we revert this to what you had before and remove the deprecation notice?  :-\

nanli-emory: reverted it. ;)

Issue #202, opened 2021-09-08
Adding Bounding Box feature


tasvora:
[openslide.PROPERTY_NAME_BOUNDS_X]

choosehappy:
i think it essentially involves this file:

https://github.com/choosehappy/HistoQC/blob/9c15c1767de0a9d1d54fa87b3e819a225b0fad00/histoqc/BaseImage.py#L138

and putting the read_region lines within an if-statement, that either the user has selected a bounding_Box = true (in the config.ini) in which case (0,0) becomes the bounding box region, or bounding_box = false in which case it says at (0,0)

Issue #220, opened 2022-09-08
Numerical precision causes slow file load


guillermojp:
Hi, 

I've just noticed that the BaseImage class is causing a problem due to openslide's inconsistencies when determining the optimal level for downsampling (see [this issue from openslide](https://github.com/openslide/openslide/issues/274)). This is especially relevant when using some functions that require 2.5x magnification. In my case, this becomes clear when checking the "level_downsamples" of the slide:

![image](https://user-images.githubusercontent.com/112631806/189143791-d2d267e8-8a5a-4a39-b59c-715a8787f158.png)

All in all, this doesn't look too bad. Issues due to numerical precision when loading the magnifications on openslide's behalf. This, however, causes a huge impact on data loading and processing in BaseImage ([line 136](https://github.com/choosehappy/HistoQC/blob/master/histoqc/BaseImage.py#L136)). Imagine I want to obtain a "2.5x" magnification. I would compute the downsampling factor, right?

![image](https://user-images.githubusercontent.com/112631806/189144722-265d8183-0a79-477c-bb2d-62f6716aff7f.png)

Ah, seems like 16. is a magnification level that is part of the slide itself (first image), corresponding to level 2 (appears as magnification of 16.00087281228769). This contrasts with the result of openslide's "get_best_level_for_downsample":

![image](https://user-images.githubusercontent.com/112631806/189145258-33b1549b-bc18-458b-a981-2b0762b37fb4.png)

In here it can be visualized better:

![image](https://user-images.githubusercontent.com/112631806/189145530-d7e7c798-6133-41c3-b13c-8c614ff7347d.png)

This seems silly but it causes the "relative_down" variable to never be exactly 1 for [line 138](https://github.com/choosehappy/HistoQC/blob/master/histoqc/BaseImage.py#L138): 

![image](https://user-images.githubusercontent.com/112631806/189145919-06cd0bc6-d720-4c6a-833d-1a5ca705ff44.png)

And even if the "level" was selected correctly, it wouldn't either be valued exactly 1:

![image](https://user-images.githubusercontent.com/112631806/189146135-95ceb13b-ba1e-48d8-8337-5ed40d4dde4b.png)

I've done a fix in the code I've been working with so that it is written as:

```python
eps = np.finfo("float16").eps # Machine epsilon of float16, eps = 0.000977
possible_levels = np.abs(np.array(osh.level_downsamples) - down_factor) < eps

if np.any(possible_levels):
[...]

PR #229, opened 2023-01-23
try to fix the issue #202 and #220:

nanli-emory:
1. wrap 'getBestThumb' to handle whether the slide has a bounding box or not
2. wrap 'getBestLevelForDownsample' to handle numerical precision for getting an appropriate level
3. wrap 'rgba2rgb' to handle merge alpha channel
4. add 'setBBox' to get the bounding box info (x, y, width, height). If there is no bounding box info, then set the bounding box as (0,0, dimensions[0], dimensions[1])
5. add 'img_bbox' property to save the bounding box info

### DIFF
--- a/histoqc/BaseImage.py
+++ b/histoqc/BaseImage.py
@@ -1,13 +1,28 @@
 import logging
 import os
 import numpy as np
-import inspect
 import zlib, dill
 from distutils.util import strtobool
+from PIL import Image
 
 #os.environ['PATH'] = 'C:\\research\\openslide\\bin' + ';' + os.environ['PATH'] #can either specify openslide bin path in PATH, or add it dynamically
 import openslide
+def getBestThumb(s, x, y, size):
+    osh = s["os_handle"]
+    if s["bounding_box"]:
+        downsample_factor = max(*(dim / thumb for dim, thumb in zip(osh.dimensions, size)))
+        level = s.getBestLevelForDownsample(downsample_factor)
+        size = tuple((np.array([s["img_bbox"][2],s["img_bbox"][3]])/osh.level_downsamples[level]).astype(int))
+        tile = osh.read_region((x, y), level, size)
+        return rgba2rgb(s, tile)
+    else:
+        return osh.get_thumbnail(size)
 
+def rgba2rgb(s, img):
+    bg_color = "#" + s["os_handle"].properties.get(openslide.PROPERTY_NAME_BACKGROUND_COLOR, "ffffff")
+    thumb = Image.new("RGB", img.size, bg_color)
+    thumb.paste(img, None, img)
+    return thumb
 
 def printMaskHelper(type, prev_mask, curr_mask):
     if type == "relative2mask":
@@ -61,7 +76,10 @@ class BaseImage(dict):
         self["dir"] = os.path.dirname(fname)
 
         self["os_handle"] = openslide.OpenSlide(fname)
-        self["image_base_size"] = self["os_handle"].dimensions
+        self["bounding_box"] = strtobool(params.get("bounding_box","False"))
+        # check if the bbox if doesn't have bbox set bounding_box to False
+        self.setBBox()
+
         self["image_work_size"] = params.get("image_work_size", "1.25x")
         self["mask_statistics"] = params.get("mask_statistics", "relative2mask")
         self["base_mag"] = getMag(self, params)
@@ -70,7 +88,7 @@ class BaseImage(dict):
         mask_statistics_types = ["relative2mask", "absolute", "relative2image"]
         if (self["mask_statistics"] not in mask_statistics_types):
             logging.error(
-                f"mask_statistic type '{self['mask_statistics']}' is not one of the 3 supported options relative2mask, absolute, relative2image!")
+                f"mask_statistic type \"{self['mask_statistics']}\" is not one of the 3 supported options relative2mask, absolute, relative2image!")
             exit()
 
         self["img_mask_use"] = np.ones(self.getImgThumb(self["image_work_size"]).shape[0:2], dtype=bool)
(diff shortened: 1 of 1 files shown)


histoqc/BaseImage.py
```diff
@@ -1,13 +1,28 @@
...
-import inspect
 import zlib, dill
 from distutils.util import strtobool
+from PIL import Image
 
 #os.environ['PATH'] = 'C:\\research\\openslide\\bin' + ';' + os.environ['PATH'] #can either specify openslide bin path in PATH, or add it dynamically
 import openslide
+def getBestThumb(s, x, y, size):
+    osh = s["os_handle"]
+    if s["bounding_box"]:
```
choosehappy: a small point here,  if the "else" code is quite small ( in this case 1 line) and includes a return statement, its better programming style to say:

```
if NOT condition
      return

-rest- of code
```

which eliminates the "else" statement entirely doesn't have a "dangling return" at the end of the function

it also reduces potential indentation errors and makes code readability/undrestanding a bit easier

histoqc/BaseImage.py
```diff
@@ -61,7 +76,10 @@ def __init__(self, fname, fname_outdir, params):
         self["dir"] = os.path.dirname(fname)
 
         self["os_handle"] = openslide.OpenSlide(fname)
-        self["image_base_size"] = self["os_handle"].dimensions
```
choosehappy, line 64: can we leave this line in for legacy purposes?

histoqc/BaseImage.py
```diff
@@ -61,7 +76,10 @@ def __init__(self, fname, fname_outdir, params):
...
+        self["bounding_box"] = strtobool(params.get("bounding_box","False"))
```
choosehappy: can you confirm that bounding_box is appearing as a column  in the output TSV file? will definitely want users to have this information in their output

histoqc/BaseImage.py
```diff
@@ -70,7 +88,7 @@ def __init__(self, fname, fname_outdir, params):
         mask_statistics_types = ["relative2mask", "absolute", "relative2image"]
         if (self["mask_statistics"] not in mask_statistics_types):
             logging.error(
-                f"mask_statistic type '{self['mask_statistics']}' is not one of the 3 supported options relative2mask, absolute, relative2image!")
+                f"mask_statistic type \"{self['mask_statistics']}\" is not one of the 3 supported options relative2mask, absolute, relative2image!")
```
choosehappy: out of curiosity, did you do this change manually or was it part of an IDE formatting? is there a reason for it or personal preference?

histoqc/BaseImage.py
choosehappy: sorry, this confused me, i see now that i was referring to img_bbox appearing in the output file.

can we modify this variable from "bounding_box" to "enable_bounding_box" or something like that, to indicate its flag variable and differentiate it from img_bbox which will contain the actual bounding box?

histoqc/BaseImage.py
```diff
@@ -90,38 +108,75 @@ def __setitem__(self, key, value):
...
+                try:
+                    self["img_bbox"] = tuple(int(num) for num in (x, y, width, height))
+                except:
+                    # no bbox info in slide set bounding_box as Flase
+                    self["bounding_box"] = False
+                    logging.warning(f"{self['filename']}: could not convert the bounding box")
+            else:
+                # no bbox info in slide set bounding_box as Flase
+                self["bounding_box"] = False
+                logging.warning(f"{self['filename']}: could not read the bounding box")
```
choosehappy: nice, i like this.....maybe we *also* put a warning again the actual image so that we can see this in the tsv file?

something like
self["warnings"].append('Bounding Box requested but could not read')

histoqc/BaseImage.py
```diff
@@ -90,38 +108,75 @@ def __setitem__(self, key, value):
...
+            y = osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_Y)
+            width = osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_WIDTH)
+            height = osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_HEIGHT)
+            if x and y and width and height:
+                try:
+                    self["img_bbox"] = tuple(int(num) for num in (x, y, width, height))
+                except:
+                    # no bbox info in slide set bounding_box as Flase
+                    self["bounding_box"] = False
+                    logging.warning(f"{self['filename']}: could not convert the bounding box")
```
choosehappy: similar note as below, but with a different message : )

histoqc/BaseImage.py
```diff
@@ -90,38 +108,75 @@ def __setitem__(self, key, value):
...
+            else:
+                # no bbox info in slide set bounding_box as Flase
+                self["bounding_box"] = False
+                logging.warning(f"{self['filename']}: could not read the bounding box")
+
```
choosehappy, line 137: this is good as it is. in practice however, I'm not sure that there is a meaningful difference on the user side between not being able to convert the bounding box and not being able to read it. if there is no difference, then this would be reduced down to a single cleaner larger try/except without an if

note as well you can use something like this:

xc=int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_X,'NA'))

which will definitely cause an error when trying to convert if it doesn't exist (this removes potential 'dangerous' of the default return value of 'get', which some libraries tools could be something like a -1 and you wouldn't detect in your code)

histoqc/BaseImage.py
```diff
@@ -1,13 +1,28 @@
...
+        tile = osh.read_region((x, y), level, size)
+        return rgba2rgb(s, tile)
+    else:
+        return osh.get_thumbnail(size)
 
+def rgba2rgb(s, img):
+    bg_color = "#" + s["os_handle"].properties.get(openslide.PROPERTY_NAME_BACKGROUND_COLOR, "ffffff")
+    thumb = Image.new("RGB", img.size, bg_color)
+    thumb.paste(img, None, img)
+    return thumb
```
choosehappy, line 29: nice! definitely was a bit lazy on my part previously : )

histoqc/BaseImage.py
```diff
@@ -131,29 +186,30 @@ def getImgThumb(self, dim):
...
 
                 target_mag = float(dim.upper().split("X")[0])
-
-                down_factor = base_mag / target_mag
-                level = osh.get_best_level_for_downsample(down_factor)
-                relative_down = down_factor / osh.level_downsamples[level]
+                downsample_factor = base_mag / target_mag
+                level = self.getBestLevelForDownsample(downsample_factor)
+                relative_down = downsample_factor/osh.level_downsamples[level]
+                size=tuple((np.array(img_base_size)/osh.level_downsamples[level]).astype(int)) if self["bounding_box"] else img_base_size
```
choosehappy: nice

choosehappy (changes requested):
Looks great! A few minor points to consider.

btw, how did you test this? did you have multiple WSIs to try it out on?

i have someone working on a project right now that uses histoqc, they can rerun the pipeline with this code modification, and i think...the results should be exactly the same if bbox = False, is that correct?

histoqc/BaseImage.py
nanli-emory: It is my personal preference. I like to keep consistent **quotation marks** (double or single) in code style/ text output style.

choosehappy (approved):

histoqc/BaseImage.py
choosehappy: i agree, but looking at the other code i think the single ones are used _in_ a print statement and a double quote is used _around_ the print statement. in this way we don't have to do escape the quotes (which i always fail to do and thus create more errors for myself : )  )

Issue #230, opened 2023-01-24
Save Macro Image module


choosehappy:
Pretty straightforward, a new module (function in saveModule file?)

it does:

```
            img=osh.associated_images["macro"]
            macroimg = np.asarray(img)[:, :, 0:3]
```

which extracts the 'macro' image taken by the scanner. note that not all scanners produce this image, so sufficient error checking will need to be in place

sometimes the image scan be quite large, so suggest also adding a option parameter for desired output image size, similar to this:

https://github.com/choosehappy/HistoQC/blob/7d11503717847edfb7d3db7c8a90556b5bfc544e/histoqc/SaveModule.py#L45

PR #231, opened 2023-01-26
Add Save Macro Image into SaveModule #230

nanli-emory:
add 'saveMarco' function to save marco image in saveModule
1. add saveAssociatedImage function to save the associated image by key
2. add saveMacro function to save the 'marco' image from slide's associated images

### DIFF
--- a/histoqc/SaveModule.py
+++ b/histoqc/SaveModule.py
@@ -36,6 +36,39 @@ def saveFinalMask(s, params):
     return
 
 
+def saveAssociatedImage(s, key:str, dim:int):
+    logging.info(f"{s['filename']} - \tsaveMacro")
+    osh = s["os_handle"]
+
+    if not key in osh.associated_images:
+        message = f"{s['filename']}- \tsave{key} Can't Read '{key}' Image from Slide's Associated Images"
+        logging.warning(message)
+        s["warnings"].append(message)
+        return
+    
+    # get asscociated image by key
+    associated_img = osh.associated_images[key]
+    (width, height)  = associated_img.size
+
+    # calulate the width or height depends on dim
+    if width > height:
+        h = round(dim * height / width)
+        size = (dim, h)
+    else:
+        w = round(dim * width / height)
+        size = (w, dim)
+    
+    associated_img = associated_img.resize(size)
+    associated_img = np.asarray(associated_img)[:, :, 0:3]
+    io.imsave(f"{s['outdir']}{os.sep}{s['filename']}_{key}.png", associated_img)
+
+def saveMacro(s, params):
+    dim = params.get("small_dim", 500)
+    saveAssociatedImage(s, "macro", dim)
+    return
+    
+
+
 def saveThumbnails(s, params):
     logging.info(f"{s['filename']} - \tsaveThumbnail")
     # we create 2 thumbnails for usage in the front end, one relatively small one, and one larger one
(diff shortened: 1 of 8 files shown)


histoqc/SaveModule.py
```diff
@@ -36,6 +36,39 @@ def saveFinalMask(s, params):
     return
 
 
+def saveAssociatedImage(s, key:str, dim:int):
+    logging.info(f"{s['filename']} - \tsaveMacro")
```
choosehappy: is this line strictly correct? this function would seem to be able to save any type of associated image. it is being called from savemacro, thus the actual image pulled is dependent upon the 'key' provided

are there other images that we will want to save, or is it only the macro which is available? if only the macro this may make more sense as a single function

i think there are two options here:
1) merge these two functions
2) saveMacro in this info needs to be - 'save{key}'? or something of the link?

histoqc/SaveModule.py
```diff
@@ -36,6 +36,39 @@ def saveFinalMask(s, params):
...
+
+    if not key in osh.associated_images:
+        message = f"{s['filename']}- \tsave{key} Can't Read '{key}' Image from Slide's Associated Images"
+        logging.warning(message)
+        s["warnings"].append(message)
+        return
+    
+    # get asscociated image by key
+    associated_img = osh.associated_images[key]
+    (width, height)  = associated_img.size
```
choosehappy, line 51: this is okay since i don't think we have a standard for histoqc, but when you look at the other tools i think you're more likely to see "nrow" and "ncol" instead of "width" and "height". just an FYI

histoqc/config/config_light.ini
```diff
@@ -18,6 +18,7 @@ steps= BasicModule.getBasicStats
 ;    BrightContrastModule.getBrightnessByChannelinColorSpace:YUV
 ;    DeconvolutionModule.separateStains
     SaveModule.saveFinalMask
+    SaveModule.saveMacro
```
choosehappy, line 21: great! 

2 more questions:
1) should we add it into the other config files (in particular think 2.1 is the new default which people should use, and use that as a template)

2) is this going to be visible/selectable from the UI? i think it just involves adding the "_macro.png" suffix somewhere in JS?

choosehappy (changes requested):

histoqc/SaveModule.py
nanli-emory: I checked some slides' associated_images. There are  'label', 'thumbnail', and 'macro' images. Let's keep **saveAssociatedImage** function just in case we need to read other associated images.

histoqc/SaveModule.py
nanli-emory: should **ncol** correspond to **height** and **nrow** correspond to **width**, right?

histoqc/config/config_light.ini
nanli-emory: I will add **SaveModule.saveFinalMask** into config.ini and add **_macro.png** suffix in JS.

histoqc/SaveModule.py
choosehappy: i think its nrow = height, ncol = width?

histoqc/SaveModule.py
choosehappy: interesting........okay good idea. maybe we can even justify adding 'label' as a save option?

PR #234, opened 2023-02-03
Refactor annotation module and handle if enable_bounding_box turns on

nanli-emory:
1. fixing the numerical precision in 'BaseImage'
2. refactoring the existing 'AnnotationModule'
3. deal with the bounding box in 'AnnotationModule'


histoqc/BaseImage.py
```diff
@@ -197,7 +197,7 @@ def getImgThumb(self, dim):
                 size=(tuple((np.array(img_base_size)/osh.level_downsamples[level]).astype(int))
                         if self["enable_bounding_box"]
                         else osh.level_dimensions[level])
-                if relative_down == 1.0: #there exists an open slide level exactly for this requested mag
+                if np.isclose(relative_down, 1, atol=.01): #there exists an open slide level exactly for this requested mag
```
choosehappy: A thought for you, perhaps "getBestLevelForDownsample" should return a binary flag as well to indicate if the level proposed is an existing openslide level?

in which case, this if statement would check the value of that boolean variable, instead of essentially doing the same computation again?

histoqc/AnnotationModule.py
```diff
@@ -95,74 +96,69 @@ def mask_out_annotation(s, point_sets):
...
-        xml_basepath = s["dir"]
-
-    xml_fname = xml_basepath + os.sep + PurePosixPath(s['filename']).stem + xml_suffix + '.xml'
-    if not Path(xml_fname).is_file():
-        msg = f"Annotation file {xml_fname} does not exist. Skipping."
+    # try use default value if the params are not provided
+    if not format:
+        msg = f"format is not provided, using xml as the default format."
+        logging.warning(f"{s['filename']} - {msg}")
+        s["warnings"].append(msg)
```
choosehappy: its not immediately clear to me how this code works? if format is None, where is it set so that the default becomes XML?

histoqc/AnnotationModule.py
```diff
@@ -95,74 +96,69 @@ def mask_out_annotation(s, point_sets):
...
+        msg = f"format is not provided, using xml as the default format."
+        logging.warning(f"{s['filename']} - {msg}")
+        s["warnings"].append(msg) 
+    
+    if not file_path:
+        # warning
+        msg = f"file path is not provided, using \"{s['dir']}\" as the default file path"
+        logging.warning(f"{s['filename']} - {msg}")
+        s["warnings"].append(msg)        
+        file_path = s["dir"]
```
choosehappy: note here that file_path is set, but in the above if statement format is not set?

choosehappy (changes requested):

histoqc/AnnotationModule.py
nanli-emory: Yes, you are right. set 'xml' as default.

histoqc/BaseImage.py
nanli-emory: Hi @choosehappy, I added a flag to indicate if the level proposed is an existing openslide level.
Also, merged the PR 236 code into BaseImage to fix the incorrect size output.

histoqc/BaseImage.py
```diff
@@ -139,89 +92,200 @@ def addToPrintList(self, name, val):
...
-                    return -1
-
-                target_mag = float(dim.upper().split("X")[0])
-                downsample_factor = base_mag / target_mag
-                level = self.getBestLevelForDownsample(downsample_factor)
-                relative_down = downsample_factor/osh.level_downsamples[level]
-                size=(tuple((np.array(img_base_size)/osh.level_downsamples[level]).astype(int))
+            return (osh.get_best_level_for_downsample(downsample_factor), False)
+
+    # this approach sequentially loads smaller tiles, resizes them, and then merges them together,
```
choosehappy: i think this comment was accidently still leftin place?

histoqc/BaseImage.py
```diff
@@ -139,89 +92,200 @@ def addToPrintList(self, name, val):
...
+                self[key] = (np.asarray(rgba2rgb(self, tile))
+                            if np.shape(tile)[-1]==4 
+                            else np.asarray(tile))        
+       
+            # specifies a desired size of thumbnail
             else:
-                logging.error(
-                    f"{self['filename']}: Unknown image level setting: {dim}!")
-                return -1
+                # recommand having the dimension is less than 10k
```
choosehappy: recommend : )

histoqc/BaseImage.py
```diff
@@ -139,89 +92,200 @@ def addToPrintList(self, name, val):
...
+            
+            win_width = int(round(win_width / cloest_downsampling_factor))
+            win_height = int(round(win_height / cloest_downsampling_factor))
+            
+            cloest_region = osh.read_region((x, y), level, (win_width, win_height))
+            if np.shape(cloest_region)[-1]==4:
+                cloest_region = rgba2rgb(self, cloest_region)
+            target_region = cloest_region.resize((win_down_width, win_down_height))
+            row_piece.append(target_region)
+        row_piece = np.concatenate(row_piece, axis=0)
```
choosehappy, line 241: can you adda comment here to note that this isn't very efficient, and if more efficiency isneeded we should likely refactor using "paste" from Image. or even just set the pixels directly with indexing?

choosehappy (changes requested):

Issue #222, opened 2022-09-21
Save mask module


choosehappy:
it would be cool to make a module called "save_mask_use" that could be placed arbitrarily within the module list in the config.ini, and whenever encountered it would simply save the current "mask_use" in memory. would need a way tag which step it was on

PR #242, opened 2023-03-21
Save mask module #222

nanli-emory:
add the saveMask function with suffix parameter and when pipeline hits the saveModule.saveMask, it will save the current "img_mask_use" as image with provided sufffix. the test config file on config_clinical.ini


choosehappy:
yup looks good!

PR #246, opened 2023-03-28
Optimized functions for speed

ajinkya-kulkarni:


histoqc/AnnotationModule.py
nanli-emory, line 1: Hi @ajinkya-kulkarni, thank you for your contribution again. Please don't let your automatic code style release our spaces.
You changed for loops to list comprehensions. Are you sure the performance get better? Do you run the code changed before and after on the same test images multiple times and compare the running time?

nanli-emory:
Close the PR since no response for several weeks

PR #248, opened 2023-04-03
fix: update function name to suppress deprecation warning

ntomita:
Addressing the following deprecation warning:

skimage_deprecation: Function ``greycoprops`` is deprecated and will be removed in version 1.0. Use ``skimage.feature.graycoprops`` instead.
  removed_version='1.0')

skimage_deprecation: Function ``greycomatrix`` is deprecated and will be removed in version 1.0. Use ``skimage.feature.graycomatrix`` instead.
  removed_version='1.0')

### DIFF
--- a/histoqc/LocalTextureEstimationModule.py
+++ b/histoqc/LocalTextureEstimationModule.py
@@ -2,7 +2,7 @@ import logging
 import numpy as np
 from skimage import  color
 from distutils.util import strtobool
-from skimage.feature import greycomatrix, greycoprops
+from skimage.feature import graycomatrix, graycoprops
 import matplotlib.pyplot as plt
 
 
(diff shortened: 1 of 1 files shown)


ntomita:
Hi @choosehappy , could you review this PR whenever you have, thanks.

nanli-emory:
> Hi @choosehappy , could you review this PR whenever you have, thanks.

Hi @ntomita, I checked this issue. we use scikit-image==0.18.1 in our requirements.txt. It shouldn't impact our code unless users installed the 0.19 and onward versions. We decide to upgrade the scikit-image to 0.19.

ntomita:
@nanli-emory Thank you for looking into it. I agree that upgrading the requirement to version 0.19 would be appropriate. Although version 0.18 works, it may be considered outdated since the latest version is already 0.20.

Issue #250, opened 2023-04-28
Will blue and black penmark be able to detect?


Cirno-tensai9:
I am using histoQC, which is quite good, but our new sample contains plenty of black and blue pen marks, I tried to add image and mask to the data/pen/ and add to the config file. 

But it always show errors when adding >=3 pen mark types

2023-04-27 17:34:04,742 - INFO - sample/TCGA-A1-A0SF-01Z-00-DX1.7F252D89-EA78-419F-A969-1B7313D77499.sv                                                          s <class 'ValueError'> Inappropriate argument value (of correct type). cannot reshape array of size 162                                                          05760 into shape (2648,2040,2)
<code object byExampleWithFeatures at 0x560dd5459720, file "~/HistoQC/h                                                          istoqc/ClassificationModule.py", line 150>

choosehappy:
Should be able to work, can you send us your data to test? the config file + masks?

CielAl:
> Should be able to work, can you send us your data to test? the config file + masks?

3972 x 2040 x 2 == 16205760, and the shape provided by the OP is (2648 x 2040 x 2)

choosehappy:
hmmm..its not clear me why there would be different sizes?

Cirno-tensai9:
The files are in this shapes and all of them can work individually:
>>> black_pen.shape
(2231, 644, 3)
>>> black_mask.shape
(2231, 644, 3)
>>> blue_pen.shape
(587, 1000, 3)
>>> blue_mask.shape
(587, 1000, 3)
>>> red_pen.shape
(1000, 918, 3)
>>> red_mask.shape
(1000, 918, 3)
>>> green_pen.shape
(853, 1000, 3)
>>> green_mask.shape
(853, 1000, 3)

I added a blue pen mark as we have a lot sampels with blue pen mark and all the other colours are as provided.

Here is one of the config files that will cause reshape error (blue+red), green+red+black will also cause similar problem
[ClassificationModule.byExampleWithFeatures:pen_markings]
name: pen_markings
threshold: .8
examples: ./pen/1k_version/pen_blue.png:./pen/1k_version/pen_blue_mask.png
          ./pen/1k_version/pen_red.png:./pen/1k_version/pen_red_mask.png

area_threshold: 100
features:  frangi
           laplace
           rgb
           #lbp
           #gabor
           #median
           #gaussian

laplace_ksize: 3

frangi_scale_range: (1,10)
frangi_scale_step: 2
[...]

PR #254, opened 2023-05-24
correct Classification Module

petroslk:
When pen_markings is used, the model used generates an output feature array of length H x W x 3. The coverslip_edge model generates an array of length H x W x 2. This output asymetry leads to an error when the "ClassificationModule.byExample:pen_markings" element of the pipeline is used in the config file because the reshape function is hardcoded to take an array of H x W x 2.

This if statement takes into account the difference in shape based on the name of the active module.

### DIFF
--- a/histoqc/ClassificationModule.py
+++ b/histoqc/ClassificationModule.py
@@ -202,7 +202,12 @@ def byExampleWithFeatures(s, params):
     img = s.getImgThumb(s["image_work_size"])
     feats = compute_features(img, params)
     cal = clf.predict_proba(feats.reshape(-1, feats.shape[2]))
-    cal = cal.reshape(img.shape[0], img.shape[1], 2)
+
+    #account for difference in output array size
+    if name == "pen_markings":
+        cal = cal.reshape(img.shape[0], img.shape[1], 3)
+    else:
+        cal = cal.reshape(img.shape[0], img.shape[1], 2)
 
     mask = cal[:, :, 1] > thresh
 


nanli-emory:
Hi @petroslk & @choosehappy,

This issue should be related to #250. Please correct me if I misunderstood anything.
The output feature array is not always H x W x n_class. It depends on the # of classes. After I read the **RandomForestClassifier.predict_proba** document:
https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier.predict_proba
I realize that the n_class should be dynamic numbers. So the reshape function should be called by using the **n_classes** in **cal.shape(n_samples, n_classes)**
I think the code should be changed to:
```python
    cal = clf.predict_proba(feats.reshape(-1, feats.shape[2]))
    # get the classes #
    cal_classes = cal.shape[1]
    # reshape by predict classes
    cal = cal.reshape(img.shape[0], img.shape[1], cal_classes)
```

This is what I thought but I'm not 100% sure.

choosehappy:
i'm confused by this... i think the error is somewhere else upstream that could be corrected so this if statement isn't needed

why is pen markings returning a  H x W x 3 and not an H x W x 2? shouldn't we *always* be doing binary classification?

nanli-emory:
Hi 

> Hi @petroslk & @choosehappy,
> 
> This issue should be related to #250. Please correct me if I misunderstood anything. The output feature array is not always H x W x n_class. It depends on the # of classes. After I read the **RandomForestClassifier.predict_proba** document: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier.predict_proba I realize that the n_class should be dynamic numbers. So the reshape function should be called by using the **n_classes** in **cal.shape(n_samples, n_classes)** I think the code should be changed to:
> 
> ```python
>     cal = clf.predict_proba(feats.reshape(-1, feats.shape[2]))
>     # get the classes #
>     cal_classes = cal.shape[1]
>     # reshape by predict classes
>     cal = cal.reshape(img.shape[0], img.shape[1], cal_classes)
> ```
> 
> This is what I thought but I'm not 100% sure.

Hi @jacksonjacobs1 , could you review the comment? The classes should be based on the type of pens. What do you think?

choosehappy:
> Hi
> 
> > Hi @petroslk & @choosehappy,
> > This issue should be related to #250. Please correct me if I misunderstood anything. The output feature array is not always H x W x n_class. It depends on the # of classes. After I read the **RandomForestClassifier.predict_proba** document: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier.predict_proba I realize that the n_class should be dynamic numbers. So the reshape function should be called by using the **n_classes** in **cal.shape(n_samples, n_classes)** I think the code should be changed to:
> > ```python
> >     cal = clf.predict_proba(feats.reshape(-1, feats.shape[2]))
> >     # get the classes #
> >     cal_classes = cal.shape[1]
> >     # reshape by predict classes
> >     cal = cal.reshape(img.shape[0], img.shape[1], cal_classes)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is what I thought but I'm not 100% sure.
> 
> Hi @jacksonjacobs1 , could you review the comment? The classes should be based on the type of pens. What do you think?

no, its a binary classification problem, either a pixel is predicted to be a pen marking, or it is not a pen marking, so the maximum dimensionality should be 2, as shown here:

https://github.com/choosehappy/HistoQC/blob/d5058ec96b3c9dcee8924853ca6c732449c9fbad/histoqc/ClassificationModule.py#L174

if the output dimension is >2, then something is wrong somewhere

does that make sense?

jacksonjacobs1:
It looks like problem is the RandomForestClassifier has 3 classes instead of 2:
```python
>>> clf.classes_
array([  0.,   1., 255.])
```

I haven't checked, but I suspect that one or more mask loaded as uint8 instead of true binary. 

Well, I kind of checked:
```python
>>> model_labels.max()
255.
```

Probably has something to do with this line: https://github.com/choosehappy/HistoQC/blob/d5058ec96b3c9dcee8924853ca6c732449c9fbad/histoqc/ClassificationModule.py#L182C59-L182C59

@nanli-emory can you look into this?

nanli-emory:
> It looks like problem is the RandomForestClassifier has 3 classes instead of 2:
> 
> ```python
> >>> clf.classes_
> array([  0.,   1., 255.])
> ```
> 
> I haven't checked, but I suspect that one or more mask loaded as uint8 instead of true binary.
> 
> Well, I kind of checked:
> 
> ```python
> >>> model_labels.max()
> 255.
> ```
> 
> Probably has something to do with this line: https://github.com/choosehappy/HistoQC/blob/d5058ec96b3c9dcee8924853ca6c732449c9fbad/histoqc/ClassificationModule.py#L182C59-L182C59
> 
> @nanli-emory can you look into this?

I will look into this. Will let you know if I fix the issue.

nanli-emory:
I create a new PR #269 for fixing the bugs

nanli-emory:
fixed this PR and bug in PR @269

Issue #252, opened 2023-05-18
Refactor BaseImage Class to support reading multiple WSI format such DICOM By using Abstract Base Class


nanli-emory:
### Description
Currently, HistoQC is using OpenSlide as WSI reader to manipulate whole-slide images. Unfortunately, OpenSlide is not really maintained anymore. It makes HistoQC only supports older file formats such svs, ndpi, big-tiff and can't handle the newer file formats such as DICOM, philips and new 3dhistec versions because of openslide.

### Issues & Potential solutions
1.  DICOM support -> use **pydicom** + **wsidicom** to create a new dicom reader 
2. the newer file formats support -> use **bioformats** to create a new bioformat reader
3. sustainable support of the latest WSI formats -> refactor BaseImage and create an abstract reader by using Abstract Base Class (ABC)
4. Bounding Box? (HistoQC support bounding box but I'm not sure if other SWI format has it or not) 

### General Class Diagram
```mermaid
classDiagram
    
    class BaseImage{
        -SWIImageReader reader
        -BoundingBox bbox
        -[int] dimensions

        -List levels
        -String magnification
        -int level_count
        -List~int~ level_downsamples
        -List~int~ level_dimensions
        +getBoundingBox()
        +GetThumbnail()
        +getTheBestThumbnail()
        +getTheBestLevelForDownsample()
        +readRegion()
    }
    class BoundingBox{
        +int x
        +int y
        +int width
        +int height

    }
    class Config{
        +bool enableBoundingBox
        +String imageWorkSize
        +enum maskStatistics
        +bool enableBoundingBox
    }
    class MaskStatistics{
        <<enumeration>>
        relative2mask
        absolute
        relative2image
    }
    class WSIImageReader~ABC~{
        <<abstract>>
        +getLevels()
        +getDimensions()
        +getLevelDimensions()
        +getLevelDownsamples()
        +getThumbnail()
        +getRegion()
    }

    class DICOMReader~WSIImageReader~{
    

    }

    class OpenSlideReader~WSIImageReader~{

    }

    class BioformatReader~WSIImageReader~{

    }
    BaseImage *-- WSIImageReader
    BaseImage *-- BoundingBox
    BaseImage *-- Config
    Config *-- MaskStatistics
    WSIImageReader <|-- DICOMReader
    WSIImageReader <|-- OpenSlideReader
    WSIImageReader <|-- BioformatReader
```    
### Readers Dependency
Reader | Libraries | Dependency
[...]

PR #256, opened 2023-06-22
first version of dicom support histoqc

nanli-emory:
#252  - Refactor BaseImage Class to support reading multiple WSI format such DICOM By using Abstract Base Class

**Add:**
1. histoqc/wsihandles/**WSIImageHandle.py** - wsi image handle (ABC) 
2. histoqc/wsihandles/**DicomHandle.py** - wsi dicom support handle
3. histoqc/wsihandles/**OpenSlideHandle.py** - openslide support handle

**Modify:**
1. histoqc/**BaseImage.py** - refactor to support and merge **WSIImageHandle**, **DicomHandle** and **OpenSlideHandle**
2. histoqc/**BasicModule.py** - change how to get the basic stats info method.
3. histoqc/**ClassificationModule.py** - fixed #254 - correct Classification Module
4. histoqc/**SaveModule.py** - change for getting macro and label image
5. histoqc/**TileExtractionModule.py** - change slide image handle from openslide to WSIImageHandle

### DIFF
--- a/histoqc/BaseImage.py
+++ b/histoqc/BaseImage.py
@@ -6,13 +6,7 @@ from distutils.util import strtobool
 from PIL import Image
 import re
 from typing import Union, Tuple
-#os.environ['PATH'] = 'C:\\research\\openslide\\bin' + ';' + os.environ['PATH'] #can either specify openslide bin path in PATH, or add it dynamically
-from histoqc._import_openslide import openslide
 from histoqc.wsihandles.WSIImageHandle import WSIImageHandle
-from histoqc.wsihandles.DicomHandle import DicomHandle
-from histoqc.wsihandles.OpenSlideHandle import OpenSlideHandle
-# there is no branch reset group in re
-# compatible with the previous definition of valid input: leading zero and leading decimals are supported
 _REGEX_MAG = r"^(\d?\.?\d*X?)"
 _PATTERN_MAG: re.Pattern = re.compile(_REGEX_MAG, flags=re.IGNORECASE)
 MAG_NA = None
@@ -37,7 +31,7 @@ class BaseImage(dict):
 
         
         # dynamically load wsi image handle
-        self["os_handle"]: WSIImageHandle = BaseImage.wsi_handle_create(fname)
+        self["os_handle"]: WSIImageHandle = WSIImageHandle.create_wsi_handle(fname)
         
         self["image_base_size"] = self["os_handle"].dimensions
         self["enable_bounding_box"] = strtobool(params.get("enable_bounding_box","False"))
@@ -109,15 +103,6 @@ class BaseImage(dict):
             return (level[0], True)
         else:
             return (osh.get_best_level_for_downsample(downsample_factor), False)
-
-
-    @staticmethod
-    def wsi_handle_create(fname: str):
-        if os.path.isdir(fname):
-            return DicomHandle(fname)
-        else:
-            return OpenSlideHandle(fname)
-
 
     @staticmethod
     def is_valid_size(size: str):
--- a/histoqc/BaseImage.py
+++ b/histoqc/BaseImage.py
@@ -30,9 +30,11 @@ class BaseImage(dict):
         self["dir"] = os.path.dirname(fname)
 
         
+        # get handles from config
+        handles = params.get("handles", "openslide,dicom")
         # dynamically load wsi image handle
-        self["os_handle"]: WSIImageHandle = WSIImageHandle.create_wsi_handle(fname)
-        
+        self["os_handle"]: WSIImageHandle = WSIImageHandle.create_wsi_handle(fname, handles)
+
         self["image_base_size"] = self["os_handle"].dimensions
         self["enable_bounding_box"] = strtobool(params.get("enable_bounding_box","False"))
         # check if the bbox if doesn't have bbox set enable_bounding_box to False
--- a/histoqc/BaseImage.py
+++ b/histoqc/BaseImage.py
@@ -8,7 +8,9 @@ import re
 from typing import Union, Tuple
 #os.environ['PATH'] = 'C:\\research\\openslide\\bin' + ';' + os.environ['PATH'] #can either specify openslide bin path in PATH, or add it dynamically
 from histoqc._import_openslide import openslide
-
+from histoqc.wsihandles.WSIImageHandle import WSIImageHandle
+from histoqc.wsihandles.DicomHandle import DicomHandle
+from histoqc.wsihandles.OpenSlideHandle import OpenSlideHandle
 # there is no branch reset group in re
 # compatible with the previous definition of valid input: leading zero and leading decimals are supported
 _REGEX_MAG = r"^(\d?\.?\d*X?)"
@@ -33,7 +35,10 @@ class BaseImage(dict):
         self["outdir"] = fname_outdir
         self["dir"] = os.path.dirname(fname)
 
-        self["os_handle"] = openslide.OpenSlide(fname)
+        
+        # dynamically load wsi image handle
+        self["os_handle"]: WSIImageHandle = BaseImage.wsi_handle_create(fname)
+        
         self["image_base_size"] = self["os_handle"].dimensions
         self["enable_bounding_box"] = strtobool(params.get("enable_bounding_box","False"))
         # check if the bbox if doesn't have bbox set enable_bounding_box to False
@@ -81,19 +86,14 @@ class BaseImage(dict):
         (dim_width, dim_height) = osh.dimensions
         self["img_bbox"] = (0, 0, dim_width, dim_height)
         # try to get bbox if bounding_box is ture
-        if self["enable_bounding_box"]:
-            # try get bbox from os handle properties
-            try:
-                x = int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_X, 'NA'))
-                y = int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_Y, 'NA'))
-                width = int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_WIDTH, 'NA'))
-                height = int(osh.properties.get(openslide.PROPERTY_NAME_BOUNDS_HEIGHT, 'NA'))
-                self["img_bbox"] = (x, y, width, height)
-            except:
-                # no bbox info in slide set enable_bounding_box as Flase
-                self["enable_bounding_box"] = False
-                logging.warning(f"{self['filename']}: Bounding Box requested but could not read")
-                self["warnings"].append("Bounding Box requested but could not read")
+        
+        # Does WSI has bounding box
+        if self["enable_bounding_box"] and osh.has_bounding_box:
+            self["img_bbox"] = osh.bounding_box
+        elif self["enable_bounding_box"] and not osh.has_bounding_box:
+            self["enable_bounding_box"] = False
+            logging.warning(f"{self['filename']}: Bounding Box requested but could not read")
+            self["warnings"].append("Bounding Box requested but could not read")        
 
     def addToPrintList(self, name, val):
         self[name] = val
@@ -109,6 +109,15 @@ class BaseImage(dict):
             return (level[0], True)
         else:
             return (osh.get_best_level_for_downsample(downsample_factor), False)
+
+
+    @staticmethod
+    def wsi_handle_create(fname: str):
+        if os.path.isdir(fname):
+            return DicomHandle(fname)
+        else:
+            return OpenSlideHandle(fname)
+
 
     @staticmethod
     def is_valid_size(size: str):
@@ -270,8 +279,7 @@ def resizeTileDownward(self, target_downsampling_factor, level):
 
 
 def rgba2rgb(s: BaseImage, img):
-    bg_color = "#" + s["os_handle"].properties.get(openslide.PROPERTY_NAME_BACKGROUND_COLOR, "ffffff")
-    thumb = Image.new("RGB", img.size, bg_color)
+    thumb = Image.new("RGB", img.size, s["os_handle"].background_color)
     thumb.paste(img, None, img)
     return thumb
 
@@ -313,22 +321,17 @@ def parsed_mag(mag: Union[str, int, float]) -> Union[None, float]:
 # this function is seperated out because in the future we hope to have automatic detection of
 # magnification if not present in open slide, and/or to confirm openslide base magnification
 def getMag(s: BaseImage, params) -> Union[float, None]:
-    logging.info(f"{s['filename']} - \tgetMag")
+    
     osh = s["os_handle"]
-    mag = osh.properties.get("openslide.objective-power") or \
-            osh.properties.get("aperio.AppMag") or MAG_NA
-    # if mag or strtobool(params.get("confirm_base_mag", "False")):
-    #     # do analysis work here
-    #     logging.warning(f"{s['filename']} - Unknown base magnification for file")
-    #     s["warnings"].append(f"{s['filename']} - Unknown base magnification for file")
-    #     return None
-    # else:
+    mag = osh.magnification or MAG_NA
     # workaround for unspecified mag -- with or without automatic detection it might be preferred to have
     # mag predefined
     mag = mag or parsed_mag(params.get("base_mag"))
     # mag is santized after invoking getMag regarding whether it's None. Therefore, it should not raise
     # the exception here.
-    return float(mag) if mag is not MAG_NA else MAG_NA
+    mag = float(mag) if mag is not MAG_NA else MAG_NA
+    logging.info(f"{s['filename']} - \tgetMag = {mag}")
+    return mag
 
 
 def getDimensionsByOneDim(s: BaseImage, dim: int) -> Tuple[int, int]:
--- a/histoqc/ClassificationModule.py
+++ b/histoqc/ClassificationModule.py
@@ -202,10 +202,7 @@ def byExampleWithFeatures(s, params):
     img = s.getImgThumb(s["image_work_size"])
     feats = compute_features(img, params)
     cal = clf.predict_proba(feats.reshape(-1, feats.shape[2]))
-    # get the classes #
-    cal_classes = cal.shape[1]
-    # reshape by predict classes
-    cal = cal.reshape(img.shape[0], img.shape[1], cal_classes)
+    cal = cal.reshape(img.shape[0], img.shape[1], 2)
 
     mask = cal[:, :, 1] > thresh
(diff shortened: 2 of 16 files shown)


histoqc/BaseImage.py
```diff
@@ -8,7 +8,9 @@
 from typing import Union, Tuple
 #os.environ['PATH'] = 'C:\\research\\openslide\\bin' + ';' + os.environ['PATH'] #can either specify openslide bin path in PATH, or add it dynamically
 from histoqc._import_openslide import openslide
-
+from histoqc.wsihandles.WSIImageHandle import WSIImageHandle
+from histoqc.wsihandles.DicomHandle import DicomHandle
```
choosehappy: should import dicomhandle and openslide handle be in this file? i would have thought they would have only been needed in the WSIImageHandle class?

histoqc/BaseImage.py
```diff
@@ -110,6 +110,15 @@ def getBestLevelForDownsample(self, downsample_factor: float) -> Tuple[int, bool
         else:
             return (osh.get_best_level_for_downsample(downsample_factor), False)
 
+
+    @staticmethod
```
choosehappy: maybe this goes in the WSI handle file?

as well, I'm wondering if these should be an if statement or a try/catch statement?

or maybe a try catch should be in place around the if? 

i guess in this case its obvious which file type to use? might need to check with petros

histoqc/ClassificationModule.py
```diff
@@ -202,7 +202,10 @@ def byExampleWithFeatures(s, params):
     img = s.getImgThumb(s["image_work_size"])
     feats = compute_features(img, params)
     cal = clf.predict_proba(feats.reshape(-1, feats.shape[2]))
-    cal = cal.reshape(img.shape[0], img.shape[1], 2)
+    # get the classes #
+    cal_classes = cal.shape[1]
+    # reshape by predict classes
+    cal = cal.reshape(img.shape[0], img.shape[1], cal_classes)
 
     mask = cal[:, :, 1] > thresh
```
choosehappy: i'm not entirely sure i understand this code here. if you have say 4 classes, won't think always produce a binary mask based on the 1st channel because of line 210? 

```
mask = cal[:, :, 1] > thresh
```

is this generalizable to multiple classes, and/or what is the situation where we have more than 2 classes? how does the user specify that which channel they want to be the defining class? 

it seems like we may only want to support 2 channel masks to keep life simple, and rely on the user to provide the correct type of input?

histoqc/SaveModule.py
```diff
@@ -62,6 +72,11 @@ def saveAssociatedImage(s, key:str, dim:int):
     associated_img = np.asarray(associated_img)[:, :, 0:3]
     io.imsave(f"{s['outdir']}{os.sep}{s['filename']}_{key}.png", associated_img)
 
+def saveLabel(s, params):
```
choosehappy, line 75: should this be refactored into a single function along with he saveMacro function? the code looks identical except for the string variable?

histoqc/wsihandles/OpenSlideHandle.py
```diff
@@ -0,0 +1,93 @@
...
+        
+        # get magnification factor from wsi slide
+        self._magnification_factor = self.osh.properties.get("openslide.objective-power") or \
+            self.osh.properties.get("aperio.AppMag")
+        
+        # get background color 
+        self._backfround_color = f"#{self.osh.properties.get(openslide.PROPERTY_NAME_BACKGROUND_COLOR, 'ffffff')}"
+
+    
+    def __get_bounding_box(self):
```
choosehappy, line 19: is it common to start private methods with two underscores instead of one? I've not seen this style before

histoqc/wsihandles/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .OpenSlideHandle import *
+from .DicomHandle import *
```
choosehappy: need to be careful with this approach as it creates a tight coupling to a new dependency, wsidicom

if wsidicom is not installed, even if the user doesn't try to use this on a DICOM image, it will still crash i think?

there are a few things I'm thinking:

1) what is the dependency size of WSIdicom? for example, if it is 10GB of installation that is one thing, if it is 100k that is another, the latter of which justifies adding in wsidicom into the requirements.txt

2) if it is a larger install, we may not want to assume the user has it, and thus only users which enable dicom support should have access to it. this can be done via e.g., the config file receiving a list of the all the handlers to use when reading WSI, this would thus extend to other reading libraries, like bioformats

3) i think one way of doing this is that there is a list created of each of the WSI handlers, in the order specified by the user, for example, in the config file

-config 
wsi_handlers= "openslidehandler","dicom","type3","type4..." 

-code
for wsi_handler in wsi_handlers:
          osh = wsi_handler.open(fname)
          if osh
                     return osh
if osh == None:
          error, no handler could read this image



notice that this logic is a bit more clean, because each handler gets to see the image (in the order specified by the user) and gets to choose if it wants to load the image or not, so it is the handler programmer that writes the code. in the version here, you see that the same concept code ("is directory" function) is one level higher, making it the main branch developers responsibility to integrate as opposed to the handler developers  job.

the interesting point is that in this manner, one need only import the libraries that the user specifies, breaking the tight coupling of having to install everything. 

for our purposes this is likely fine its only 2 handlers, and the dependencies are minimal, but for a larger scale project with more developers, or more handlers, the way above is definitely more correct. 

do you see the difference?  perhaps you want to take a stab at refactoring that way for your own understanding, i think it shouldb't be more than a couple minutes of effort

choosehappy (changes requested):

histoqc/BaseImage.py
nanli-emory: I can't import DicomHandle and OpenSlideHandle in WSIImageHandle class since DicomHandle and OpenSlideHandle need to import WSIImageHandle as the superclass. It causes an error if I import DicomHandle and OpenSlideHandle in WSIImageHandle class:
```ImportError: cannot import name 'WSIImageHandle' from partially initialized module 'histoqc.wsihandles.WSIImageHandle' (most likely due to a circular import)```

File: histoqc/BaseImage.py (hunk 2 above)
Comment (nanli-emory) [reply histoqc/BaseImage.py]: I still think about this part. We probably should use factory pattern + dynamic import to implement how to create a subclass for ABC class. 
Do you have any thoughts on this?  @CielAl

File: histoqc/SaveModule.py (hunk 4 above)
Comment (nanli-emory) [reply histoqc/SaveModule.py:75]: Actually, we have a single function named **saveAssociatedImage** with the string parameter. The **saveLabel** and **saveMacro** functions are restrictions of using **saveAssociatedImage** since wsidicom and openslide image and APIs are heterogeneous. In openslide, **associated_images** property is _**Images, such as label or macro images, which are associated with this slide. This is a Mapping from the image name to RGBA Image.**_ We can directly get the associated image by the key such as **openslide.associated_images["macro"]**. However, in wsidicom, there is no associated image property. The wsidicom only provides 2 functions to get associated images - **WsiDicom.read_label()** and **WsiDicom.read_overview()**. 
Saving label and macro functions should be the most frequently used functions. This is why I create these 2 functions individually. Open to any better proposal. ;)

File: histoqc/wsihandles/OpenSlideHandle.py (hunk 5 above)
Comment (nanli-emory) [reply histoqc/wsihandles/OpenSlideHandle.py:19]: Yes. Please see the following test code for the private functions.
```python
class Student:
    def _not_private(self):  # not private function
	    print('This is not private function.')
                 
    def __private(self): # private function
	    print('This is private function.')

std = Student()
# call no private function
std._not_private()

# Using name mangling to call private function
std._Student__private()

# AttributeError
std.__private() 
```
But some people said _Python's **private** functions are not actually private since user can use name mangling to call private function_
reference: https://www.geeksforgeeks.org/private-methods-in-python/

histoqc/BaseImage.py
CielAl: @nanli-emory @choosehappy 
I agree.
If I understand correctly about what you want here:
(1) avoid importing each subclass handles because not all might be installed on the client's machine.
(2) dynamically create the corresponding subclass handle based on what's installed or any input arguments.

So herein you may have the below options:
(1) create a factory class (or make it extending Protocol for Python 3.10 and above) that implements the relevant APIs and the builder:
`
    
class HandleBuilder:

    @staticmethod
    def build(params) -> WSIImageHandle:

        ... # return different handle types based on 

`

(2) Optionally, rather than just making a HandleBuilder, make the supposed factory builder a higher level of abstraction in some way close to the skeletal pattern. Herein your WSIImageHandle would be the abstract class to manage how to open the image resources and compute corresponding attributes by the data read accordingly.
You can create another interface (e.g., `class ImageCore` )that 
    --- (1)  associate a WSIImageHandle as a private attribute.
    --- (2) provide all APIs of WSIImageHandle by returning the results of the associated WSIImageHandle 
    --- (3) implement the factory builder function above as a static method or class method

By doing this the advantage is, you can move certain behavior like "get_best_level_for_downsample" from BaseImage to `ImageCore` and therefore your BaseImage can focus on carrying data/mask/results between each module. 

(3) Simply use a function as the factory builder.
image

histoqc/ClassificationModule.py
nanli-emory: Hi @choosehappy,
It is relevant to #250 and #254. My explanation is on #254.

histoqc/BaseImage.py
CielAl: @nanli-emory @choosehappy 
Note that essentially you want to hide the details of all WSIImageHandle's subclasses from BaseImage as HistoQC's pipeline itself shouldn't care about it anyway, therefore the import of all subclasses should be avoided in the BaseImage class. 

The purpose of the second proposal is essentially to organize the BaseImage class so that all WSI-related methods that should not be in a dict subclass (BaseImage) can be moved to this layer. I am personally kinda wary of extending a python built-in type and lump too many overrides there. 

In some way, the BaseImage at the moment is deeply coupled to all modules and most of the modules access BaseImage by string keys -- note IDEs usually have limited functionality to refactor string literals, so even if you fully annotated these keys by Literal, it will be annoying to manage future changes.

With this abstraction, you can fold the dict of BaseImage as an associated attribute, refactor the corresponding helper function of compression, and integrate all corresponding key-value pairs as the unified interface:
e.g. you can make the BaseImage the abstraction I mentioned but associate the actual dict as an attribute besides the WSIImageHandle,
`
class BaseImage:


    __base_image_dict: dict
    __wsi_image_handle:

 
    # build from dict of params
    def __init__(self, params):
         self.__wsi_image_handle = some-factory-builder
         self.__base_image_dict = another-builder-for-init-baseimagedict-from-params


`
and therefore with IDE it's easier to refactor the Modules for future changes and can make fewer mistakes (e.g., typos of keys).

histoqc/ClassificationModule.py
choosehappy: left a comment there. i still don't understand why this is needed or why we're not always doing binary classification? shouldn't need any multi-class stuff

histoqc/SaveModule.py
choosehappy: i think you can remove both of these functions, and have them be keywords used directly in "saveAssociatedImage" via the config file?

you basically already have this, except instead of having "key" come in as a method-level parameter, you read it from the config file?

histoqc/wsihandles/OpenSlideHandle.py
choosehappy: super cool, i didn't know about this functionality, thanks!

histoqc/wsihandles/__init__.py
choosehappy: did you see this comment as well?

histoqc/BaseImage.py
choosehappy: "Note that essentially you want to hide the details of all WSIImageHandle's subclasses from BaseImage as HistoQC's pipeline itself shouldn't care about it anyway, therefore the import of all subclasses should be avoided in the BaseImage class."

Yes, exactly this. this is what it looks like, which also addresses the first comment in this review and last comment:

```python
#---main.py
from baseclass import AbstracthbaseClass

user_list=["openslide","dicom"]

for types in user_list:
    ci = AbstracthbaseClass.returnClass(types)
    print(ci)
    print(ci.has_bounding_box)
from abc import ABC, abstractmethod

#---baseclass.py
class AbstracthbaseClass(ABC):


    @staticmethod
    def returnClass(type):
        if type=="openslide":
            print("c1")
            from c1class import c1
            return c1(type)
        else:
            print("c2")
            from c2class import c2
            return c2(type)

    @property
    @abstractmethod
    def has_bounding_box(self):
        ...

#--- c1class.py
from baseclass import AbstracthbaseClass

class c1(AbstracthbaseClass):
    def __init__(self, fname):
        self.fname = fname
        print("c1 init")

    @property
    def has_bounding_box(self):
        print("c1 bounding box")
        return self.fnamefrom baseclass import AbstracthbaseClass

#----c2class.py
class c2(AbstracthbaseClass):
    def __init__(self, fname):
        self.fname = fname
        print("c2 init")

    @property
    def has_bounding_box(self):
        print("c2 bounding box")

```

nanli-emory: Hi @choosehappy, I Changed how to load and create subclasses for WSIImageHandle. But do we really need to have "handles" parameters for ImageBase? 
Currently, **create_wsi_handle** function will auto-determine the file type and import the subclass if needed.

histoqc/wsihandles/WSIImageHandle.py
```diff
@@ -1,7 +1,17 @@
...
+from pathlib import Path
+import os, logging
+
+
+class WSIHandles(Enum):
+    OPENSLIDE = 'openslide'
+    WSIDICOM = 'dicom'
+
+class WSIFileExtension(Enum):
+    OPENSLIDE = ['.bif', '.mrxs', '.ndpi', '.scn', '.svs', '.svslide', '.tif', '.tiff', '.vms', '.vmu']
```
choosehappy: where did you get this list of file extensions? just want to make sure its exhaustive

histoqc/wsihandles/WSIImageHandle.py
```diff
@@ -93,4 +103,53 @@ def read_label(self):
...
+            logging.error(msg)
+            raise NotImplementedError(msg)
+    
+def is_dicom(path: Union[str, Path]) -> bool:
+    # Return rrue if the file at `path` is a DICOM dir.
+
+    # is dicom file
+    def __is_dicom(f):
+        with open(f, 'rb') as fp:
+            fp.read(128)  # preamble
```
choosehappy: i've not seen this before, do you have a reference you can link to or something similar? trying to understand how robust this may be

histoqc/wsihandles/__init__.py
choosehappy: @nanli-emory , please see this comment and confirm that you saw it?

histoqc/wsihandles/WSIImageHandle.py
```diff
@@ -93,4 +103,53 @@ def read_label(self):
...
     def read_macro(self):
-        ...
\ No newline at end of file
+        ...
+
+    
+    @staticmethod
+    def create_wsi_handle(fname):
+        # determine the file type
+        if is_dicom(fname):
```
choosehappy: i think you may be trying too hard here, you're trying to come up with a narrow definition of what a dicom is, and I'm not sure that that is going to work, e.g., i know that all of them don't end in dcm, and it is highly dependent upon the manufacturer

as well, the user should know about their environment

i would suggest modifying this to be a try-catch situation based on the readers the user has in their config. i feel like i made this comment before, did you have some concern about it? i just pointed it out again, can you please confirm that you saw it?

lets think about this a little deeper. what do we ultimately want? the handler to try to and open the file, right?
what is wrong with something like this:

```python
for reader in config_reader_list
	import reader
	try:
		osh = reader.open(fname)
	exception:
		pass
if osh == None:
	error -- no reader wants to read this file
	
```

where now the *reader itself* gets to determine if it wants to open the file, as opposed to your approach here which requires both a sophisticated algorithm to see if it is a dicom, and as well a list of file extensions

ultimately, if the reader can open it, its a success, so we can use that directly as a success criteria?

since the user is providing the list of readers that are set up in their system, we need not worry about importing things which they don't think are enabled

thoughts?

choosehappy (changes requested):
Getting there! requesting better refactoring to simplify reader selection

histoqc/wsihandles/WSIImageHandle.py
nanli-emory: Hi @choosehappy I got it from [openslide website](https://github.com/choosehappy/HistoQC/pull/256).

histoqc/wsihandles/WSIImageHandle.py
nanli-emory: Hi @choosehappy, I saw this on pydicom code. https://github.com/pydicom/pydicom/blob/d21ecad8e663c8937d064da561bf7299415d7d98/pydicom/misc.py#L39

histoqc/wsihandles/WSIImageHandle.py
nanli-emory: Hi @choosehappy  I move the create_wsi_handle into WSIImageHandle.

histoqc/wsihandles/WSIImageHandle.py
choosehappy: got it, thanks. i still think we should move away from this approach as discussed in my github review, did you see it?

histoqc/wsihandles/WSIImageHandle.py
choosehappy: is this comment intended in response to what i said above?

histoqc/wsihandles/WSIImageHandle.py
CielAl: It's probably not ideal to identify which reader to use by extension, especially when what these readers can handle are not mutually exclusive.

You might want to define a default option if nothing is specified or whatever library requested by the user is not installed, e.g., import these readers by a specified order and use whatever is correctly imported. Otherwise, it should be defined by the user, e.g., put a string or int placeholder in the config file to specify what type of WSIHandle it is.

nanli-emory:
hi @choosehappy,
can I merge this PR if there are no more code changes as required?

choosehappy:
lets wait for some feedback from petros first

requirements.txt
```diff
@@ -11,4 +11,5 @@ typing-extensions
 
 requests~=2.22.0
 Pillow~=8.4.0
```
petroslk, line 13: bumping up to Pillow~=9.1.0 should solve dependency conflicts with wsidicom

PR #259, opened 2023-07-23
Rework AnnotationModule.py and misc

### LINKED ISSUE #258: AnnotationModule did not exclude holes from GeoJSON's polygon and multipolygons
Per the standard of GeoJSON (example here: https://gist.github.com/andrewharvey/971b9db1900a62efa7635f6a6d337ae7#file-polygon-with-a-hole-geojson), for a `Polygon` geometry type, only the first vertices list represents the outer ring while the rest are holes.

However, the current implementation considers all inner rings as individual polygons - and the corresponding pixel regions are set to 1 in `mask_out_annotation`. Naturally, these operations are also redundant since all holes are definitely inclusive inside the outer ring: https://github.com/choosehappy/HistoQC/blob/master/histoqc/AnnotationModule.py#L68

I recommend re-model the template point-set functions to return Tuple[point_set,  List[point_set]] therefore the mask_out_annotation can correctly identify holes.

I doubt the ImageScope's xml has a well-defined behavior to assign holes into a specific node/attribute so perhaps we can ignore that part?

CielAl:
Fix #258 
@nanli-emory @choosehappy
(1) Rework AnnotationModule.py (adding histoqc.annotations module- dependency: shapely which requires numpy>=1.14 - compatible). 

- no changes in config files/outputs
- shapely is added mainly because it's a better abstraction of geometries that can handle exterior and interior rings. It also supports spatial indexing trees if we ever want to introduce fast region selection by annotation and masks in the future.
- use PIL.ImageDraw.polygon instead of skimage.draw.polygon as the former directly works on a provided mask handle, therefore hiding internal loops to fill pixel values. 

(2) Move dynamical imports to a standalone package (openslide and typing related 
- perhaps we can move forward from python3.7~python3.8 to python3.8+ in the future? More typing annotation supports and syntactic sugars could make life easier, besides many other AIML libraries are extended to new major Python versions.

(3) Add other required Windows binaries/lib for openslide into bin/. If we provide a precompiled dll of openslide in the local folder, all its dependencies should also be present. The binaries and the import wrapper of openslide are basically bandaids to mitigate the change of behavior of dll imports in Windows.


requirements.txt
```diff
@@ -1,14 +1,17 @@
...
-Pillow~=8.4.0
-setuptools~=49.2.1
\ No newline at end of file
+lazy-property
+requests~=2.28.1
+Pillow~=9.4.0
+setuptools~=65.6.3
+shapely~=2.0.1
+zlib~=1.2.13
+openslide~=20171122
```
nanli-emory: Is this only for Windows build?

requirements.txt
```diff
@@ -1,14 +1,17 @@
...
-requests~=2.22.0
-Pillow~=8.4.0
-setuptools~=49.2.1
\ No newline at end of file
+lazy-property
+requests~=2.28.1
+Pillow~=9.4.0
+setuptools~=65.6.3
+shapely~=2.0.1
+zlib~=1.2.13
```
nanli-emory: Is this only for Windows build?

requirements.txt
CielAl: nah it's probably the autofill of requirments.txt via IDE when I code in Windows. The shapely works for linux and windows.

nanli-emory: Please remove `openslide~=20171122` from the requirements.txt. I think it is only for windows and will have a error if install the requirements.txt on my macOS. Thanks

requirements.txt
CielAl: The requirements.txt was autofilled since we moved the windows binaries of openslide into the bin folder.

Note that the `libopenslide-0.dll` depends on other binaries in the win build, e.g., zlibs, and therefore if we want to provide user a usable binary that can be properly imported, we should put all relevant files to the bin rather than just `libopenslide-0.dll` in the previous version.

Corresponding entries were removed.

nanli-emory (approved):

PR #269, opened 2023-11-07
Correct classification module

nanli-emory:
fixed correct Classification Module #254
1. convert masks to binary if they are grayscale masks
2. add warning logs if the masks are not binary images


jacksonjacobs1 (approved):
Looks good!

PR #272, opened 2023-11-22
small deconvolution bug fix

jacksonjacobs1:


histoqc/DeconvolutionModule.py
```diff
@@ -23,9 +23,9 @@ def separateStains(s, params):
         sys.exit(1)
         return
 
-    stain_matrix = getattr(sys.modules[__name__], stain, "")
+    stain_matrix = getattr(sys.modules[__name__], stain, None)
 
-    if stain_matrix == "":
+    if stain_matrix is None:
```
choosehappy, line 28: i think this is more pythonic as 

```
if not stain_matrix::
```

jacksonjacobs1:
```python
if not stain_matrix:
```
This line would work if stain_matrix were a python array, but not for an ndarray. That's because numpy uses their own Truthy behavior:
```python
>>> import numpy as np
>>> arr = np.array([1,2,3])
>>> if arr:
...     print("something")
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

choosehappy: got it, what about this however:

```
>>> arr = np.array([])
>>> if arr:
...     print("something")
...
<stdin>:1: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
```

should we use arr.size?

```
>>> if arr.size:
...     print("test")
... else:
...     print("no")
...
```

but i guess then this would need to be changed to:
```
stain_matrix = getattr(sys.modules[__name__], stain, np.array([]))
```

probably starting to overengineer this : )

jacksonjacobs1: Would you consider checking for an empty array more pythonic than "is None"? 

I like NoneType here because there's less ambiguity. Otherwise, an empty array could mean the getattr() used the default value OR the stain attribute actually stored an empty array.

Thoughts?

choosehappy: yea, i think you got it. None seems like the better option here!

jacksonjacobs1: As well, I think this is the behavior we want, right?

```python
>>> import numpy as np
>>> arr = np.array([])
>>> if arr is None:
...     print("array is none")
... else:
...     print("array is not none")
... 
array is not none

```

PR #290, opened 2024-04-25
Full multiprocessing GPU support.

CielAl:
@nanli-emory 

…Make CuImage Handle prototype ready for future development. Massive code cleaning

Briefly: 
(1) The WSIImageHandle abstraction is ready to use. OpenSlideHandle is deployed. 
CuImageHandle is implemented but is not enabled in image handle factory as more time is needed for a unified downstream computational interface in custom modules.
(2) Code cleaning including forward porting which prepares for the future deprecated function signatures in skimage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ wheels/
 *.egg-info/
 *.egg
 *.whl
+*.log
 .coverage*
 MANIFEST
 histoqc/tests/data/*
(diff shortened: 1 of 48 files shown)


CielAl:
Update the modules to automatically call the GPU-accelerated operations given the device of WSIImageHandle.

Issue #276, opened 2023-12-15
# of pieces of tissue - feature


choosehappy:
from mask to use count + report # of unique pieces of tissue

should help identify e.g. TMAs and number of spots on the TMA

CielAl:
Seems like a simple regionprop + minimum area thresholding can work. Just return the length of `skimage.measure.regionprops` output list and that's it.

CielAl:
Question would be - should that be in the final processing stage, or that the # of tissue pieces should be traced at the end of each module (as post processing). IMO reporting in the final processing is sufficient.

choosehappy:
agree, final post processing is likely going to be sufficient

as well, I'm not sure that we have to make a hard decision

if we define the module, users can stick the module where they want in the pipeline, and even use it multiple times if they give it an associated tag via a prefix like this one:

https://github.com/choosehappy/HistoQC/blob/72d1093f748e6a0bf55cd9212eefcaa936183e0f/histoqc/config/config_v2.1.ini#L163

jacksonjacobs1:
@nanli-emory , can you add this to your list?

HistoTools project item:
https://github.com/users/jacksonjacobs1/projects/2/views/1?pane=issue&itemId=62031151&sortedBy%5Bdirection%5D=desc&sortedBy%5BcolumnId%5D=Status

We will basically just need a function called `count_tissue_pieces()` which gets the count as @CielAl described and records the count in the results tsv.

The question is, do we need a new module for this function or can we stick it in e.g, `MorphologyModule.py` and add it to the end of the v2.1 pipeline?

CielAl:
@jacksonjacobs1 @nanli-emory 
I think a standalone module might be overkilling. A few facts here helping us wrapping it into existing modules:
(1) Multiple modules involve collecting regionprops and corresponding stats (e.g., max/min of areas) and this piece of code is duplicated a lot. The # of tissue area is in fact also implicated in this computation.
(2) ```regionprops``` is pickleable.

Therefore it might helpful to wrap regionprop-related statistics into a helper function wherein all modules can use it directly: e.g., given a mask (or regionprops, if already computed) and returns the # of connected tissue area, the summary statistics of these areas, etc. The resulted regionprops can be saved into the BaseImage as it's pickleable.

The advantage is that it reduces the amount of duplicated codes involving regionprops, and it also provides flexibility to report # of connected tissue areas after any arbitrary module (e.g., how many tissue pieces left after module1 and how many after module2, etc.).

PR #296, opened 2024-05-08
get num tissues

nanli-emory:
feature added: # of pieces of tissue [#276](https://github.com/choosehappy/HistoQC/issues/276)

1.  add helper function `BaseImage.getMaskReginsStats` to get the based statistics for tissue areas
2. Use `BasicModule.countTissuePieces` to get # tissues in the current mask
3. change the `./histoqc/config/config_v2.1.ini` files to save `#pieces_of_tissue`


histoqc/BasicModule.py
```diff
@@ -69,3 +69,10 @@ def finalProcessingArea(s, params):
             f"{s['filename']} - After BasicModule.finalProcessingArea NO tissue remains detectable! Downstream modules likely to be incorrect/fail")
         s["warnings"].append(
             f"After BasicModule.finalProcessingArea NO tissue remains detectable! Downstream modules likely to be incorrect/fail")
+
+
+def countTissuePieces(s, params):
+    area_thresh = int(params.get("area_threshold", "1000"))
```
jacksonjacobs1: I think applying an area threshold here is redundant. Ideally small objects etc. will already be filtered out and the `countTissuePieces` method will just report what's there.

histoqc/BlurDetectionModule.py
```diff
@@ -37,19 +37,20 @@ def identifyBlurryRegions(s, params):
...
-        nobj = area_max = area_mean = 0
+    # rps = measure.regionprops(morphology.label(mask))
+    # if rps:
+    #     areas = np.asarray([rp.area for rp in rps])
+    #     nobj = len(rps)
+    #     area_max = areas.max()
+    #     area_mean = areas.mean()
+    # else:
+    #     nobj = area_max = area_mean = 0
+    stat_rs = getMaskReginsStats(mask)
```
jacksonjacobs1: Looks like none of the stats are being added to the printList?

histoqc/BaseImage.py
```diff
@@ -341,3 +342,22 @@ def getDimensionsByOneDim(s: BaseImage, dim: int) -> Tuple[int, int]:
     else:
         w = int(dim * width / height)
         return w, dim
+
+
+def getMaskReginsStats(img, area_threshold: int = 0)-> dict:
```
jacksonjacobs1: regions misspelled

histoqc/BaseImage.py
```diff
@@ -206,6 +206,9 @@ def getImgThumb(self, size: str):
                 self[key] = getBestThumb(self, bx, by, target_dims, target_sampling_factor)
         return self[key]
 
+'''
+the followings are helper functions
```
jacksonjacobs1: "following" not "followings"

jacksonjacobs1:
Thanks Nan, 

Just a few small comments above.

nanli-emory:
> Thanks Nan,
> 
> Just a few small comments above.

Hi @jacksonjacobs1, please review and confirm the changes.

jacksonjacobs1 (approved):
Looks good! Ready to merge in.

jacksonjacobs1:
@choosehappy one quick question before I merge this in....

This PR adds a line to config v2.1 to count # of tissue pieces.

I do not think we should modify config files while keeping the same filename. Conceptually, histoqc should always produce the same output if the user uses config v2.1. Do you agree?

choosehappy:
that is a very good, even philosophical, question

at minimum I would say the same config file should produce the same numerical results, and in this case, simply counting the pieces of tissue isn't actually impacting the *metrics*

on the other hand, the output file itself now changes - which is less than ideal ,but may be tolerable

how do you want to address this? i think we need config files to be "fully featureD" so that folks can see how to use the different things. on the other hand, "2.1" was for the 2.1 release --- are we still on the 2.1 release if we keep adding things? 

at some point we should officially release (tag + code) e,g. 2.2, 2.3 etc - maybe having a config name based on the version is a bit chaotic and unsustainable? maybe it should all go into just config.ini and have the versioning "do its thing"? 

what do you think?

jacksonjacobs1:
I agree that the config file v2.1 naming convention is a bit chaotic.

However, pushing changes into the config.ini kind of defeats the purpose of nan's unit tests:
https://github.com/choosehappy/HistoQC/blob/master/histoqc/tests/test_images_tsv_results.py

The unit tests were written specifically for a configuration file (in this case v2.1) which we assume does not change between commits.

If we change config files in commits, we will also need to update the unit tests before each commit.

choosehappy:
> If we change config files in commits, we will also need to update the unit tests before each commit.

I don't immediately know that this is a bad thing? if we're adding new features, we should be adding new unit tests to cover those features? 

so it somehow seems like this is the intended outcome?

PR #301, opened 2024-05-16
added debug option

### LINKED ISSUE #300: Add "debug" mode & set matplotlib backend based on mode
Agg should be set as the matplotlib backend by default instead of TKAgg. 

Add a `--debug` option (argparse store_true) which sets the backend to TKAgg when passed. The `--debug` option should also allow for other debugging functionality to be toggled if needed.

jacksonjacobs1:
fixes #300


jacksonjacobs1:
The debug flag now changes the debugging level. @choosehappy can you take a look at this?

choosehappy:
looks good!

PR #323, opened 2025-07-24
add option to output pixels_to_use as a percentage of the total numbe…

### LINKED ISSUE #322: Calculation of pixels to use as percent
It would be great to add another calculation for the percent of total (original height and width) for what % passed QC.  This then enables an easy pass/fail criterion to be made for routine work.  We worked out the calculation and taking the base (40x for Leica) and the working magnification (usually 1.25x) you could then work out a needed divisor of 32.  So then you take the pixels to use / (multiply height x width / 32) and convert to percent and voila, good number to use.  Thanks!
Michael

jacksonjacobs1:
…r of pixels in the image

### DIFF
--- a/histoqc/BasicModule.py
+++ b/histoqc/BasicModule.py
@@ -22,7 +22,7 @@ def getBasicStats(s, params):
 
 def finalComputations(s, params):
     mask = s["img_mask_use"]
-    s.addToPrintList("pixels_to_use", str(len(mask.nonzero()[0])))
+    s.addToPrintList("pixels_to_use", printMaskHelper(params.get("mask_statistics", s["mask_statistics"]), mask, mask))
 
 
 def finalProcessingSpur(s, params):
--- a/histoqc/config/config.ini
+++ b/histoqc/config/config.ini
@@ -168,6 +168,8 @@ blur_threshold: .15
 
 
 [BasicModule.finalComputations]
+; two options: absolute, relative2image. relative2mask is not available here
+mask_statistics = absolute
 
 [BasicModule.finalProcessingSpur]
 disk_radius: 5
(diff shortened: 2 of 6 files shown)


jacksonjacobs1:
I've confirmed that the updated functionality works as intended and can be merged in.

Note that this change does not change the default behavior of HistoQC. To output `pixels_to_use` as a percentage rather than a pixel count, modify the following line within any given config file:

Default
```ini
[BasicModule.finalComputations]
; two options: absolute, relative2image. relative2mask is not available here
mask_statistics = absolute
```

Modified
```ini
[BasicModule.finalComputations]
; two options: absolute, relative2image. relative2mask is not available here
mask_statistics = relative2image
```

histoqc/BasicModule.py
```diff
@@ -22,7 +22,7 @@ def getBasicStats(s, params):
 
 def finalComputations(s, params):
     mask = s["img_mask_use"]
-    s.addToPrintList("pixels_to_use", str(len(mask.nonzero()[0])))
+    s.addToPrintList("pixels_to_use", printMaskHelper(params.get("mask_statistics", s["mask_statistics"]), mask, mask))
```
choosehappy, line 25: would probably add a logger.warning here if the value is "relative2mask" stating that this is not available here and thus the value is being set to zero (and maybe a warning in the output tsv as well)? don't want to end up in the situation where folks are using things that they don't realize?

choosehappy (commented):
take a look at this comment plllllllease

histoqc/BasicModule.py
jacksonjacobs1: Sure, done here: https://github.com/choosehappy/HistoQC/pull/327

PR #326, opened 2025-07-24
export all filtered rows into TSV file, including rows without comments

### LINKED ISSUE #320: Add export feature
Hi, 

I was wondering if you could add a feature to export filtered images as some kind of table again?

 Best,

jacksonjacobs1:


jacksonjacobs1:
If filtering has been applied to the dataset, only filtered rows will be exported into the `results_modified.tsv` file.

PR #328, opened 2025-10-08
Main

fsumayyamohamed:


jacksonjacobs1:
HIi @fsumayyamohamed ,

Can you describe why this PR is necessary? It looks like you've modified parameters as instructed here:
https://scikit-image.org/docs/0.23.x/release_notes/release_0.20.html

However, the version of scikit image in the histoqc requirements.txt file is 0.19, not 0.20: https://github.com/choosehappy/HistoQC/blob/f2137e97281b21e2effa2837574a120c0157c64d/requirements.txt#L2

PR #333, opened 2025-11-20
Task/190-add-output-folder-structure-to-histoqc-docs

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/190

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -16,4 +16,4 @@ wiki/index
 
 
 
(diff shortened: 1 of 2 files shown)


jacksonjacobs1:
@choosehappy do you have any other suggestions to add to the histoqc documentation?

For instance, I think we agreed that the base_mag feature needed to be documented.

choosehappy:
yes to the base_mag

there is also nothing ray specific in here, is there? that would seem to be the same part that needs to be added from the previous version?

